PR #4, opened 2024-02-16
Utilities for multivariate gaussians

richfitz:
Some utilities for working with multivariate gaussians:

* compute the density
* compute the derivative of the density with respect to the parameters (assuming zero mean)

Currently merging into #1 because it contains those commits due to the support for sampling from a multivariate gaussian


R/distributions.R
```diff
@@ -13,3 +13,31 @@ make_rmvnorm <- function(vcv) {
     x + drop(rnorm(n) %*% r)
   }
 }
+
+
+ldmvnorm <- function(x, vcv) {
```
M-Kusumgar, line 18:
```suggestion
# log density multivariate normal
ldmvnorm <- function(x, vcv) {
```

R/distributions.R
```diff
@@ -13,3 +13,31 @@ make_rmvnorm <- function(vcv) {
...
+  dec <- base::chol(vcv)
+  constant <- -sum(log(diag(dec))) - 0.5 * nrow(vcv) * log(2 * pi)
+  function(x) {
+    constant - 0.5 * sum(backsolve(dec, x, transpose = TRUE)^2)
+  }
+}
+
+
+## Not the world's most beautiful name.
+dldmvnormdx <- function(x, vcv) {
```
M-Kusumgar, line 33: perhaps?
```suggestion
deriv_ldmvnorm <- function(x, vcv) {
```

R/distributions.R
```diff
@@ -13,3 +13,31 @@ make_rmvnorm <- function(vcv) {
...
+  make_dldmvnormdx(vcv)(x)
+}
+
+
+make_dldmvnormdx <- function(vcv) {
```
M-Kusumgar, line 38: similarly
```suggestion
make_deriv_ldmvnorm <- function(vcv) {
```

tests/testthat/test-distributions.R
```diff
@@ -29,3 +29,27 @@ test_that("rvnorm produces correct expectation", {
...
   expect_equal(var(r), vcv, tolerance = 0.01)
 })
+
+
+test_that("can compute the log density of a multivariate normal", {
+  vcv <- matrix(c(4, 2, 2, 3), ncol = 2)
+  set.seed(1)
+  x <- matrix(rnorm(20, sd = 10), ncol = 2)
+  expect_equal(apply(x, 1, make_ldmvnorm(vcv)),
+               mvtnorm::dmvnorm(x, sigma = vcv, log = TRUE))
```
M-Kusumgar, line 39: is our implementation of the `rmvnorm` better? wondering why we implemented that in the distributions file

M-Kusumgar (changes requested):
Looks good, would be nice if we had a little bit more explanation on the variable names!

tests/testthat/test-distributions.R
richfitz: Ours caches the decomposition. Later ours will also be translated into C++

M-Kusumgar (approved):
LGTM :)

PR #5, opened 2024-02-16
Basic HMC implementation

richfitz:
Basic HMC implementation. 

~Follows from #1 and #4 and #7~
Follows from #10


R/sampler-hmc.R
```diff
@@ -0,0 +1,222 @@
...
+      }
+      if (debug) {
+        history[i + 1, ] <- pars_next
+      }
+    }
+
+    ## Make a half step for momentum at the end.
+    v_next <- v_next + epsilon * compute_gradient(pars_next) / 2
+
+    ## Negate momentum at end of trajectory to make the proposal symmetric
```
MJomaba, line 84: The negative change can just be dropped here as vm is independently resampled at each step.

richfitz:
Not sure why actions are not running here...

MJomaba (approved):
Code is looking good. Tests on simple Gaussian and banana shaped distribution working well.

PR #6, opened 2024-02-20
Import dust rng headers, with name translation

richfitz:
Avoiding the temptation to do any refactoring yet, though some will come.  This is the RNG and distribution support from dust, with just the simple substitution of the word 'dust' for 'mcstate' everywhere.

~Against #1, which I would really like to get merged soon.~

Genuinely unreviewable, but this is exactly the implementation and tests from dust, refined over many PRs and considerable use there.


M-Kusumgar (approved):
Thanks for the explanation, going to approve since this is just a copy and paste job for now!

PR #7, opened 2024-02-22
Use our RNG for everything

richfitz:
This PR avoids use of R's RNGs everywhere, using the ones introduced in #6 instead.  This will be used in another PR (#8) which allows running in parallel while ensuring that results are identical regardless of the number of processes they run on.

This does introduce a performance regression, but we'll sort that out later. There are a bunch of small interface changes that it will be nice to make to the handling of RNGs from R-space; they were never really intended to be used this way!  We also have some scope after this PR to write a sampler that runs in C++ code and which we can parallelise with openmp, which will be pretty quick I think.


tests/testthat/test-sampler-random-walk.R
```diff
@@ -26,18 +17,9 @@ test_that("validate construction of rw sampler", {
...
-    shape <- 1
-    rate <- 1
-    mcstate_model(
-      parameters = "gamma",
-      direct_sample = function() rgamma(1, shape = shape, rate = rate),
-      density = function(x) dgamma(x, shape = shape, rate = rate, log = TRUE),
-      gradient = function(x) (shape - 1) / x - shape,
-      domain = rbind(c(0, Inf)))
-  })
+  m <- ex_simple_gamma1()
```
weshinsley, line 20: Makes no difference to test passing - 
but should this be `model`, not `m` for lines 26/28?

weshinsley (changes requested):
One uninteresting thing on  test at the end...

R/sample.R
```diff
@@ -68,7 +72,16 @@ mcstate_sample <- function(model, sampler, n_steps, initial = NULL) {
   colnames(history_pars) <- model$parameters
 
   ## I'm not sure about the best name for this
-  details <- sampler$finalise(state, model)
+  details <- sampler$finalise(state, model, rng)
+
+  if (!identical(get_r_rng_state(), r_rng_state)) {
+    cli::cli_warn(c(
+      "Detected use of R's random number generators",
+      i = paste("Your model has used R's random number generators (e.g.,",
```
M-Kusumgar, line 80:
```suggestion
      i = paste("Your model has used R's random number generators (e.g.",
```

tests/testthat/test-distributions.R
```diff
@@ -1,20 +1,20 @@
...
   vcv <- matrix(c(4, 2, 2, 3), ncol = 2)
   x <- runif(4)
-  set.seed(1)
-  r1 <- rmvnorm(x, vcv)
-  set.seed(1)
-  r2 <- make_rmvnorm(vcv)(x)
+  g1 <- mcstate_rng$new(seed = 42)
+  g2 <- mcstate_rng$new(seed = 42)
+  r1 <- rmvnorm(x, vcv, g1)
+  r2 <- make_rmvnorm(vcv)(x, g2)
```
M-Kusumgar, line 7: I feel like this is testing two perpendicular things, them being: `g1` and `g2` the same and `rmvnorm(arg1, arg2, arg3)` and `make_rmvnorm(arg2)(arg1, arg3)` the same but i think thats alright

M-Kusumgar (approved):
sorry for the delay, happy with this PR, had tiny comments but all are optional

R/sample.R
richfitz: https://en.wiktionary.org/wiki/e.g. - e.g. is usually followed by a comma :)

R/distributions.R
```diff
@@ -1,16 +1,15 @@
...
 
 
-##' @importFrom stats rnorm
 make_rmvnorm <- function(vcv) {
   n <- ncol(vcv)
   r <- chol(vcv, pivot = TRUE)
   r <- r[, order(attr(r, "pivot", exact = TRUE))]
-  function(x) {
-    x + drop(rnorm(n) %*% r)
+  function(x, rng) {
```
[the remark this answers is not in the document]
richfitz: hmm, I'm not sure on this, but I can see where you're coming from here.

The reason I don't want to change it is with this version we will write at the point of use:

```
rmvnorm(x, rng)
```

Later we'll probably move the multivariate normal support within the rng class, at which point we'd write


```
rng$rmvnorm(x)
```

so this preserves the connection with `rng` and makes it clear we have the same RNG generator.  

We might also use the transformation captured by the closure here with any number of rng objects, perhaps in parallel (this will become more obvious in #8, I think when there are many `rng` objects floating around).  The way we have it here we capture `vcv`, which will be the same everwhere, but *not* the `rng` which can be different.

tests/testthat/test-sampler-random-walk.R
richfitz: yup, fixed for clarity.  the reason why it never works is because we never force the promise (we fail before `model` was ever referenced, as in:

```
f <- function(x) 1
f(whatever)
```

R/sample.R
M-Kusumgar: damn, i had no idea :open_mouth:

weshinsley (approved):

R/distributions.R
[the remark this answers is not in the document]
EmmaLRussell: This makes sense to me, I think, especially if it's going to end up in the rng class anyway. I suppose you could make rng an optional parameter on `rmvnorm` and if it's not provided then use a default rng, to allow for simpler usage. But maybe better to make the client think about appropriate rng usage for the scenario.

R/sample.R
```diff
@@ -68,7 +72,16 @@ mcstate_sample <- function(model, sampler, n_steps, initial = NULL) {
...
+                "via rnorm, runif, sample, etc).  This means that your",
+                "results will not be reproducible as you change the sample",
+                "runner")))
```
EmmaLRussell, line 83: Can we suggest what they should do instead? Point them at rng?

tests/testthat/test-distributions.R
```diff
@@ -23,15 +23,16 @@ test_that("rmvnorm has correct behaviour in trivial case", {
...
-  set.seed(1)
-  r <- t(replicate(1e5, f(x)))
+  r <- t(replicate(1e5, f(x, rng)))
   expect_equal(colMeans(r), c(1, 2), tolerance = 0.01)
   expect_equal(var(r), vcv, tolerance = 0.01)
 })
 
 
 test_that("can compute the log density of a multivariate normal", {
+  skip_if_not_installed("mvtnorm")
```
EmmaLRussell, line 35: What's the idea with these skips? If we're running tests either manually or on CI, don't we want to run all of them, and install what's needed?

richfitz: Long story.  

* The package is only a "suggested" dependency (it's only used in tests) so this means that the test won't run if it's not available, which is what CRAN requires
* By default, the CI checks won't even start (the build will error almost immediately) if suggested packages are not found so we don't ever really see this.  However, it will be nice on (say) a train if the package is not there, which is what prompted me to remember to add this
* More annoyingly, CRAN will sometimes run checks with suggested packages missing, in order to send emails to people requiring that they apply this sort of fix, in order to complain they are overburdened.

EmmaLRussell: oh CRAN... 😆

EmmaLRussell (approved):

PR #8, opened 2024-02-22
Allow running more than one chain, with configurable runners

richfitz:
This PR allows us to run more than one chain at once.  I've set this up via a simple "chain runner" which can be serial or parallel at the moment.  There's not really much to either method and neither are suitable for production use as there's no progress reporting, which is not ok for something that might happily run for hours.

For the runner, at the moment I've just returned a list with a `run` function and a class attribute.  We might want to add more to this in future, but if not we can roll back to a function with a class if we want.  I do wonder if we might want some sort of initialisation / run / finalisation split, but because we never reuse the object this seems less worthwhile.  Where we are most likely to change in future is if we have runner that can report progress; in that case the most general solution is probably:

* have the runner advertise that it has this functionality
* add a poll function that reports on the progress over all chains
* have the actual sample function look after hitting that repeatedly and rendering some progress output

Because this builds off of #7, the random number draws are identical between both the serial and parallel runners, and this will be easy to ensure for future runners too.  It would be fairly easy to have the rng state updated back into the rng object after each chain, but we don't currently return that anyway so I've not done this.

Most of the pain here is in sorting out initial parameter states for all the chains; that's pulled into one gory function that is the same for everything; we start with a matrix of parameters for all runners.

~Merge after #7, contains those commits.  Currently pointing at that branch so that the diff is easy to see.~


tests/testthat/helper-mcstate2.R
```diff
@@ -1,8 +1,15 @@
 ex_simple_gamma1 <- function(shape = 1, rate = 1) {
-  mcstate_model(
-    parameters = "gamma",
-    direct_sample = function(rng) rng$gamma(1, shape = shape, scale = 1 / rate),
-    density = function(x) dgamma(x, shape = shape, rate = rate, log = TRUE),
-    gradient = function(x) (shape - 1) / x - rate,
-    domain = rbind(c(0, Inf)))
+  e <- new.env(parent = .GlobalEnv)
```
richfitz, line 2: This bit of butchery won't be obvious, but is in order to prevent a warning that "namespace mcstate may not be available on load" while serialising for parallel use.

R/sample.R
```diff
@@ -97,3 +103,100 @@ mcstate_sampler <- function(name, initialise, step, finalise) {
...
+  initial
+}
+
+
+combine_chains <- function(res) {
+  stopifnot(is.null(res$chain))
+  chain <- rep(seq_along(res), lengths(lapply(res, "[[", "density")))
+
+  pars <- rbind_list(lapply(res, "[[", "pars"))
+  density <- unlist(lapply(res, "[[", "density"))
```
M-Kusumgar, line 176: just for clarification, why is pars a matrix and density a vector here? is it because we expect multiple parameters in the pars so making it a matrix is better but density will be `1 x n` where `n` is number of samples?

R/sample.R
```diff
@@ -34,25 +44,23 @@ mcstate_sample <- function(model, sampler, n_steps, initial = NULL) {
...
-    pars <- initial
-    if (length(pars) != length(model$parameters)) {
-      cli::cli_abort(
-        paste("Unexpected initial parameter length {length(pars)};",
-              "expected {length(model$parameters)}"),
-        arg = "initial")
-    }
-  }
 
+mcstate_run_chain <- function(pars, model, sampler, n_steps, rng) {
```
M-Kusumgar, line 62: I think this function should be in the `runner.R` file, nothing in this file references this

R/runner.R
```diff
@@ -0,0 +1,94 @@
...
+    ## "package X not being available", and also a chance that objects
+    ## will not survive serialisation/deserialisation between
+    ## processes here - but we'll face that on basically every
+    ## possible parallel backend really.
+    ##
+    ## It would be nice to refactor the rng so that we are using
+    ## something closer to the rng pointer object rather than the big
+    ## heavy R6 object as that already has the logic in for doing
+    ## state sync, but this can all be done transparently later.
+    pars_list <- lapply(seq_len(nrow(pars)), function(i) pars[i, ])
```
M-Kusumgar, line 75: not sure if its faster, just a suggestion :P
```suggestion
    pars_list <- asplit(pars, MARGIN = 1)
```

R/sample.R
```diff
@@ -97,3 +103,100 @@ mcstate_sampler <- function(name, initialise, step, finalise) {
...
 }
+
+
+initial_parameters <- function(initial, model, rng, call = NULL) {
+  n_pars <- length(model$parameters)
+  n_chains <- length(rng)
+  if (is.null(initial)) {
+    ## Really this would just be from the prior; we can't directly
+    ## sample from the posterior!
+    initial <- lapply(rng, function(r) model$direct_sample(r))
```
M-Kusumgar, line 114: i am assuming this is not guaranteed to be able to take in a vector?

M-Kusumgar (changes requested):
Looks good to me! took a little bit longer to review this one, couple of tiny comments

R/util.R
```diff
@@ -39,3 +39,25 @@ get_r_rng_state <- function() {
...
+}
+
+
+rbind_list <- function(x) {
+  stopifnot(all(vlapply(x, is.matrix)))
+  if (length(x) == 1) {
+    return(x[[1]])
+  }
+  nc <- unique(vnapply(x, ncol))
+  nr <- vnapply(x, nrow)
```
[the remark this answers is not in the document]
richfitz: no, just a leftover. I was going to do something more complicated then decided to do it this way for now at least. Have removed the `nr`

R/sample.R
```diff
@@ -97,3 +103,100 @@ mcstate_sampler <- function(name, initialise, step, finalise) {
...
+                    "element{?s}")),
+        arg = "initial", call = call)
+    }
+    err <- lengths(initial) != n_pars
+    if (any(err)) {
+      ## We might want to highlight the entries that are bad, and also
+      ## indicate if it was direct_sample() that produced the invalid
+      ## entry.
+      cli::cli_abort(
+        c(paste("Unexpected initial parameter length; expected {n_pars}"),
```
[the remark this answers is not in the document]
richfitz: ugh, was just being lazy.  I've added something here but room to improve it remains for later

R/sample.R
richfitz: density is a mapping from pars -> real, with pars on R^n.  So in the history we have n_steps densities but pars is a n_steps x length(pars) matrix.

R/runner.R
richfitz: TIL

R/sample.R
richfitz: I don't understand the questions here (or which bit of this is not taking in a vector)

R/runner.R
```diff
@@ -0,0 +1,130 @@
...
+##' It has nothing that can be configured (yet).
+##'
+##' @title Run MCMC chain in series
+##'
+##' @return A runner of class `mcstate_runner` that can be passed to
+##'   [mcstate_sample()]
+##'
+##' @export
+mcstate_runner_serial <- function() {
+  run <- function(pars, model, sampler, n_steps, rng) {
```
EmmaLRussell, line 12: As you know, I got confused about what exactly rng needed to be here, and how that defined the number of chains, so maybe a comment on that would be useful, unless it's going to be really obvious.

R/runner.R
```diff
@@ -0,0 +1,130 @@
...
+    on.exit(parallel::stopCluster(cl))
+    ## There is a good chance the user will see warnings about
+    ## "package X not being available", and also a chance that objects
+    ## will not survive serialisation/deserialisation between
+    ## processes here - but we'll face that on basically every
+    ## possible parallel backend really.
+    ##
+    ## It would be nice to refactor the rng so that we are using
+    ## something closer to the rng pointer object rather than the big
+    ## heavy R6 object as that already has the logic in for doing
```
EmmaLRussell, line 73: Is the heaviness of the R6 object the reason why we pull the state out of rng to pass to the parallel cluster, rather than passing it directly?

R/runner.R
```diff
@@ -0,0 +1,130 @@
...
+
+## Later we could return the mutated rng state and set it back into
+## the sampler, if we needed to continue for any reason.
+mcstate_run_chain_parallel <- function(pars, model, sampler, n_steps, rng) {
+  rng <- mcstate_rng$new(rng)
+  mcstate_run_chain(pars, model, sampler, n_steps, rng)
+}
+
+
+mcstate_run_chain <- function(pars, model, sampler, n_steps, rng) {
```
EmmaLRussell, line 97: This has just been moved from another file?

R/sample.R
```diff
@@ -34,47 +44,107 @@ mcstate_sample <- function(model, sampler, n_steps, initial = NULL) {
...
+  ret <- list(name = name,
+              initialise = initialise,
+              step = step,
+              finalise = finalise)
+  class(ret) <- "mcstate_sampler"
+  ret
+}
+
+
+initial_parameters <- function(initial, model, rng, call = NULL) {
```
EmmaLRussell, line 72: It looks like this method will check initial parameters we generate from `model$direct_sample` as well as user-provided ones, which perhaps we don't need to do if we trust ourselves (but maybe we don't). Should we check for silly values like NULL (?) as well as for the right number of values?

R/sample.R
```diff
@@ -34,47 +44,107 @@ mcstate_sample <- function(model, sampler, n_steps, initial = NULL) {
...
-    state <- sampler$step(state, model, rng)
-    history_pars[i + 1, ] <- state$pars
-    history_density[[i + 1]] <- state$density
-  }
 
-  ## Pop the parameter names on last
-  colnames(history_pars) <- model$parameters
+combine_chains <- function(res) {
+  stopifnot(is.null(res$chain))
+  chain <- rep(seq_along(res), lengths(lapply(res, "[[", "density")))
```
EmmaLRussell, line 139: Sorry, struggling to parse what this gets you. The lengths of each chain?

R/util.R
```diff
@@ -39,3 +39,24 @@ get_r_rng_state <- function() {
...
+
+
+vlapply <- function(...) {
+  vapply(..., FUN.VALUE = TRUE)
+}
+
+
+vnapply <- function(...) {
+  vapply(..., FUN.VALUE = 1)
+}
```
EmmaLRussell, line 51: I had to look these up to work out what was going on here!

EmmaLRussell (commented):
More questions than anything useful!

R/runner.R
richfitz: No, here it's because we can't serialise the object because the pointer gets nulled, so we send the state through instead

R/sample.R
richfitz: a sequence 1, 1, 1, 1..., 2, 2, 2, 2, .... etc (1 * n_steps, 2 * n_steps, ..., n_chains * n_steps)

R/sample.R
M-Kusumgar: ahh dw, was talking about doing `model$direct_sample(rng)` instead of the lapply but yh i dont think that will always work

R/sample.R
EmmaLRussell: Oh, so it tells you which chain each density and pars value comes from?

M-Kusumgar (approved):

R/sample.R
richfitz: yeah exactly that

R/runner.R
richfitz: sorry this reply is on entirely the wrong question!

R/runner.R
richfitz: No, here it's because we can't serialise the object because the pointer gets nulled, so we send the state through instead

R/runner.R
richfitz: Mantra suggested moving this bit here, yeah

R/sample.R
richfitz: > we generate from model$direct_sample as well as user-provided ones

parameters from `direct_sample` are also really user defined as they've written that function and there's no guarantee that these are the right length.  `NULL` is covered by a check for length, though we could add a numeric check in too

R/runner.R
EmmaLRussell: I thought it probably was! :laughing:

R/sample.R
EmmaLRussell: Ah yeah, that's true!

EmmaLRussell (approved):

weshinsley (approved):
Cool - looks good to me - one question on the way that I think I already know the answer to!

R/runner.R
```diff
@@ -0,0 +1,130 @@
...
+    rng_state <- lapply(rng, function(r) r$state())
+
+    res <- parallel::clusterMap(
+      cl,
+      mcstate_run_chain_parallel,
+      pars = pars_list,
+      rng = rng_state,
+      MoreArgs = list(model = model, sampler = sampler, n_steps = n_steps))
+  }
+  structure(list(run = run), class = "mcstate_runner")
```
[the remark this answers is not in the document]
richfitz: we'll need to do that I think yes, but probably this won't survive the PoC stage as we'll need to also get things like progress reporting in here

PR #9, opened 2024-02-23
Support for stochastic models

richfitz:
Note that this PR pulls in `dust` for now, to use in the tests.  That is probably not what we really want because dust is going to depend on mcstate for its rng calls!  Once we work out what we need, I'll hand-write a version of this model that does the job, but this setup should be ok for a while (probably until we know enough to write dust2, really).

I've left a little placeholder in around *getting* rng state from the model; that is going to interact with restartable mcmc in ways that are going to require some care I think.

There's no actual testing in the stochastic model + random walk; however, this constitutes "particle MCMC" which is cool.  There are a couple of small features we will likely add there later based the covid work:

* the ability to rerun the likelihood every so often separate from the proposal, avoiding getting stochastically stuck (I'll properly work through this with Marc and Ed)
* the ability to indicate the minimum likelihood that we'd accept (this turns out to be useful for getting an early exit in the particle filter)

### DIFF
--- a/tests/testthat/test-sampler-random-walk.R
+++ b/tests/testthat/test-sampler-random-walk.R
@@ -38,7 +38,7 @@ test_that("can draw samples from a random model", {
   vcv <- matrix(c(0.0006405, 0.0005628, 0.0005628, 0.0006641), 2, 2)
   sampler <- mcstate_sampler_random_walk(vcv = vcv)
   res <- mcstate_sample(m, sampler, 20)
-  expect_setequal(names(res), c("pars", "density", "details"))
+  expect_setequal(names(res), c("pars", "density", "details", "chain"))
   ## OK, this is ready then to start thinking about how we inject
   ## observer support in here; we'll need access to the actual model
   ## in order to get trajectories out etc, but that's not super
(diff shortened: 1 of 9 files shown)


tests/testthat/test-sampler-random-walk.R
```diff
@@ -30,3 +30,18 @@ test_that("validate sampler against model on initialisation", {
...
+  vcv <- matrix(c(0.0006405, 0.0005628, 0.0005628, 0.0006641), 2, 2)
+  sampler <- mcstate_sampler_random_walk(vcv = vcv)
+  res <- mcstate_sample(m, sampler, 20)
+  expect_setequal(names(res), c("pars", "density", "details", "chain"))
+  ## OK, this is ready then to start thinking about how we inject
+  ## observer support in here; we'll need access to the actual model
+  ## in order to get trajectories out etc, but that's not super
+  ## obvious really; that means we're going to need some sort of
+  ## special sauce for the observer function really.
+})
```
weshinsley, line 47: I have not understood this comment well yet!

weshinsley (approved):
OK - I think I understand what the PR is adding in terms of the stochastic support - but I haven't grasped the final comment about observers etc - so perhaps missing a bit of bigger picture here... but other than that...!

tests/testthat/test-sampler-random-walk.R
richfitz: ah, it's out of date now after #10

richfitz: (and now it's gone)

M-Kusumgar: would like to definitely chat through this and that sir model on call!

M-Kusumgar (commented):
Looks good! just some minor comments

M-Kusumgar (approved):

PR #10, opened 2024-02-26
Improve interface around model specification

richfitz:
To support having an observer on a stochastic model that can save things like trajectories, and to support adding lots of weird things to support nested models, I think we need a slightly more flexible approach to specifying models.

The motivation for this is a little hard to see, but starts to appear when considering models that run a time series as part of the likelihood calculation. For these, we're going to want to record trajectories over time (either deterministic or some sampled stochastic trajectories).  We're likely also to want to store things like the final state so that onward simulation can be continued.

The `mcstate_model` interface does not need to know about any of this - and we don't want to have an ever increasing set of things to add there as people add new functionality to their models.  Instead I'm imagining that we'll have some sort of "observer" that will be added to the call to `mcstate_sample` that we'll run at each step and which will receive the mcstate model to prod at it however it wants.  So the observer and the model need to agree on the interface but the sampler does not really need to know about this detail at all.

Two ways we could do this are to allow `mcstate_model` to accept `...` with additional properties, or we can compose the final `mcstate_model` out of whatever we're given; I've gone with the latter.

~I was not totally sure where to put `parameters` (and then as a result `domain`) so for now I've put them _both_ in the model and in the call to `mcstate_model` but I'm open to removing one of these and requiring that we have these in one place or the other.  I'd wondered too about allowing `model` to be a density function (in which case we just do `model <- list(density = density)` to allow a really simple interface here; this is much of where allowing `parameters`/`domain` through has come from. For stochastic models (#9, currently draft) having this format is useful because it allows a model to provide functions that can support a stochastic model and then also gives us something to use to say "but actually this one is deterministic", which is what we do when creating deterministic models from stochastic ones~

Edit: @M-Kusumgar and I had a chat about this and we're gone for something slightly different here that ends up being a little easier to think about I think.  There's now an `mcstate_model_properties` function which can be passed in to `mcstate_model` which let's you assert what your model can and cannot do.  We then look at the model and pull out all the functions that we need into a list but we also keep the original model which allows us to have the user do whatever they want at each mcmc step (in a later PR)


R/model.R
```diff
@@ -1,63 +1,235 @@
...
+##' * `domain`: Information on the parameter domain.  This can be
+##'   overridden by the `domain` argument to this function.  This is a
+##'   two column matrix with `length(parameters)` rows representing
+##'   each parameter.  The parameter minimum and maximum bounds are
+##'   given as the first and second column.  Infinite values (`-Inf`
+##'   or `Inf`) should be used where the parameter has infinite domain
+##'   up or down.  Currently used to translate from a bounded to
+##'   unbounded space for HMC, but we might also use this for
+##'   reflecting proposals in MCMC too.  If not present (and if not
+##'   overridden by `domain`) we assume that the model is valid
```
M-Kusumgar, line 60:
```suggestion
##'   reflecting proposals in MCMC too.  If not present we assume that the model is valid
```

R/model.R
```diff
@@ -1,63 +1,235 @@
...
+##'   density.  It must take an argument representing a parameter
+##'   vector (a numeric vector) and return a single value.  This is
+##'   the posterior probability density in Bayesian inference, but it
+##'   could be anything really.  Models can return `-Inf` if things
+##'   are impossible, and we'll try and cope gracefully with that
+##'   wherever possible.
 ##'
-##' @param direct_sample A function to sample directly from the
+##' * `parameters`: A character vector of parameter names.  This can
+##'   be overridden by the `parameters` argument to this function.
```
M-Kusumgar, line 47:
```suggestion
##' * `parameters`: A character vector of parameter names. 
```

R/model.R
```diff
@@ -1,63 +1,235 @@
...
+##'   gradient, where these are called with the same parameters.  This
+##'   function is optional (and may not be well defined or possible to
+##'   define).
 ##'
-##' @return An object of class `mcstate_model`, which can be used with
-##'   a sampler.
+##' @title Create basic model
+##'
+##' @param model A list or environment with elements as described in
+##'   Details.
```
M-Kusumgar, line 85: where is "Details"? is this a roxygen2 thing that will reference the text above?

M-Kusumgar (approved):
Looks great! lets see how this propagates through the other PRs but i like it, couple tiny documentation comments

R/model.R
richfitz: Everything after the first linebreak is Details in the help rendering (or should be)

R/model.R
```diff
@@ -1,63 +1,232 @@
...
-##' directly.  Contains data and methods that define a basic model
-##' object, so that we can implement samplers against.  Not all models
-##' will support everything here, and we'll add additional
-##' fields/traits over time to advertise what a model can do.  For
-##' example, models will need to advertise that they are capable of
-##' being differentiated, or that they are stochastic in order to be
-##' used with different methods.
+##' Describe properties of a model.  Use of this function is optional,
+##' but you can pass the return value of this as the `properties`
+##' argument to enforce that your model does actually have these
```
EmmaLRussell, line 3:
```suggestion
##' argument of mcstate_model to enforce that your model does actually have these
```

R/model.R
```diff
@@ -1,63 +1,232 @@
...
+            "your properties say that it should do"),
+      arg = "model", call = call)
+  }
+  if (!is.null(direct_sample) && !is.function(direct_sample)) {
+    cli::cli_abort(
+      "Expected 'model$direct_sample' to be a function if non-NULL",
+      arg = "model", call = call)
+  }
+  direct_sample
+}
```
EmmaLRussell, line 221: Could this repeated logic be implemented once and take expected property name etc as argument?

richfitz: yeah, I wondered about that. Will do

weshinsley (approved):

PR #11, opened 2024-02-29
Banana

MJomaba:
* Adding the banana shaped distribution as an example model
* Add a simple test on that with HMC sampler
* Create a simple doc describing the example, to be used later as a vignette, with comparison of the random walk and HMC samplers


richfitz:
Thanks @MJomaba; I've pulled the banana into #5 for now so we'll get that merged with the HMC implementation and then I'll pull in the Rmd changes once we set up some basic docs

PR #12, opened 2024-03-01
Implement restartable chains

richfitz:
This PR implements restartable MCMC chains, with the requirement that a chain run in two chunks will be exactly the same as a chain run in one go.  This is going to require some additional thought when it comes in contact with some other in-flight work (which I anticipate doing in another PR perhaps, or depending on how merges go we could do it here):

* #9: we'll need to dump state from the model and reload.  This requires us to know (sort of) if the model is capturing the rng object or seeding from that object.  This will require in some cases ability to get the state back out of the model.
* #3 and #5: we need to provide an alternative to `initialise` for stateful samplers so that they can dump out their internal state and reload it.

I've left the idea of being restartable behind an off-by-default option, because it obliges us to retain a copy of the modek, runner and sampler used in the sampling.  That might be ok, but it will certainly bulk out the objects, particularly when they're saved to disk (and they may not all even serialise generally).  I'm not sure that this is all that bad though really.

I've also not pursued the idea that one might want to continue while changing the sampler or the runner

* the runner should always be easy to change because we will make sure that it won't change the results.  So we could offer a `runner` option that overrides the one baked in with relatively little increase in complexity
* the sampler _could_ be changeable but then we're doing something slightly different to a restart and are really just starting a new chain from the end point of an old one.  In particular there's no guarantee that appending two chains together is a reasonable thing to do


R/sample.R
```diff
@@ -145,6 +201,7 @@ combine_chains <- function(res) {
   density <- unlist(lapply(res, "[[", "density"))
   details <- lapply(res, "[[", "details")
   details <- if (all(vlapply(details, is.null))) NULL else details
+  rng_state <- lapply(res, function(x) x$internal$rng_state)
```
weshinsley, line 204: Should this get used yet?

R/sample.R
```diff
@@ -156,16 +213,62 @@ combine_chains <- function(res) {
...
+  samples <- list(pars = pars,
+                  density = density,
+                  details = details,
+                  chain = chain)
+  class(samples) <- "mcstate_samples"
+  samples
+}
+
+
+## This is absolutely terrible, but it will get there.
```
weshinsley, line 225: Yeah - the next few lines I am somewhat treating as a blackbox, but the tests are very clear and working!

tests/testthat/test-sample.R
```diff
@@ -146,3 +146,63 @@ test_that("need a direct sample function in order to start sampling", {
...
+                         restartable = TRUE)
+
+  set.seed(1)
+  res2a <- mcstate_sample(model, sampler, 10, 1, n_chains = 3,
+                          restartable = TRUE)
+  res2b <- mcstate_sample_continue(res2a, 10, restartable = TRUE)
+  res2c <- mcstate_sample_continue(res2b, 10, restartable = TRUE)
+
+  expect_equal(res2c, res1)
+})
```
weshinsley, line 182: So is the difference between these two tests just the extra res2c? Which we would not be able to do had we not included restartable=TRUE for res2b

weshinsley (approved):
OK - think I can understand this one

R/sample.R
```diff
@@ -156,16 +213,62 @@ combine_chains <- function(res) {
...
+append_chains <- function(prev, curr) {
+  n_chains <- length(prev$restart$rng_state)
+  i <- split(seq_along(prev$chain), factor(prev$chain, seq_len(n_chains)))
+  j <- lapply(
+    split(seq_along(curr$chain), factor(curr$chain, seq_len(n_chains))),
+    function(x) x[-1] + length(prev$chain))
+  k <- unlist(matrix(c(i, j), 2, byrow = TRUE))
```
M-Kusumgar, line 232: completely optional but i think it's less gross, also why are we skipping that first step of the restarted chain? is it because that'll be the initial so it will be repeated?
```suggestion
  i <- split(seq_along(prev$chain), prev$chain)
  j <- split(seq_along(curr$chain) + length(prev$chain), curr$chain)
  j <- lapply(j, function(x) x[-1])
  k <- unlist(rbind(i, j))
```

M-Kusumgar (approved):
This is quite a nice PR, enjoyed it :P, just a small suggestion and comment

R/sample.R
richfitz: > why are we skipping that first step of the restarted chain? is it because that'll be the initial so it will be repeated

Yes - exactly that. Ed and I are discussing if the first step is really part of the chain anyway; we might decide to discard it an retain initial conditions instead, in which case this will simplify slightly

R/sample.R
richfitz: vestigial I think, from before I tidied up the restart data

tests/testthat/test-sample.R
richfitz: yeah, it tests that we can restart a chain that was restarted; so both that we get the same answer as the combo chain (`res1`) and that the restart information is copied over

R/sample.R
```diff
@@ -156,16 +212,61 @@ combine_chains <- function(res) {
...
+}
+
+
+## This is absolutely terrible, but it will get there.
+append_chains <- function(prev, curr) {
+  n_chains <- length(prev$restart$rng_state)
+  i <- split(seq_along(prev$chain), prev$chain)
+  j <- split(seq_along(curr$chain) + length(prev$chain), curr$chain)
+  j <- lapply(j, function(x) x[-1])
+  k <- unlist(rbind(i, j))
```
EmmaLRussell, line 230: This is just to get the the `chain` chain indexes lined up nicely for the appended chains?

EmmaLRussell (approved):
Makes sense as far as I can tell!

PR #13, opened 2024-03-01
Basic HMC implementation (take 2)

richfitz:
Basic HMC implementation.

~Follows from #1 and #4, #7 and #10~ 
Was originally #5 but I was unable to resurrect that PR

Marc has reviewed already from a statistical perspective.


R/sampler-hmc.R
```diff
@@ -0,0 +1,220 @@
...
+hmc_transform <- function(domain) {
+  lower <- domain[, 1]
+  upper <- domain[, 2]
+
+  is_bounded <- is.finite(lower) & is.finite(upper)
+  is_semi_infinite <- is.finite(lower) & upper == Inf
+  is_infinite <- lower == -Inf & upper == Inf
+  unknown <- !(is_bounded | is_semi_infinite | is_infinite)
+  if (any(unknown)) {
+    ## This is not very likely to be thrown
```
weshinsley, line 143: So this would be... 
lower == upper == (-Inf)
lower == upper == (Inf)
or
lower = -Inf and upper is finite   ? (Or is this semi-infinite?)

R/sampler-hmc.R
```diff
@@ -0,0 +1,220 @@
...
+
+    ## Derivative of rn2model, as a multiplier to f'(x); we pass in
+    ## model parameters here, not R^n space.
+    deriv = function(x) {
+      ## d/dtheta theta -> 1
+      x[is_infinite] <- 1
+      ## Here we want
+      ## d/dtheta f(exp(theta)) -> exp(theta) f'(exp(theta))
+      ##                        -> x f'(x)
+      x[is_semi_infinite] <- x[is_semi_infinite]
```
weshinsley, line 174: Not sure I see what this line means

weshinsley (approved):
Ok - I think I understand how this fits together, although bit off the HMC I need more thinking on! Two comments/things I was a bit confused on

R/sampler-hmc.R
richfitz: `(-Inf, 0]` would error here, yeah.  But I don't know of any distributions like that!  We can support that if needed fairly simply but no point complicating things until we need to

R/sampler-hmc.R
richfitz: that line can be removed, but is in there to remind me that's what we're doing really! Probably this would be clearer if we assigned into something other than `x`; I'll tweak that

R/sampler-hmc.R
```diff
@@ -0,0 +1,220 @@
...
+## Substituting the theta -> x transformation log((x - a) / (b - x)) into this
+##
+## > (b - a) exp(log((x - a) / (b - x))) / (exp(log((x - a) / (b - x))) + 1)^2
+## > (a - x) * (b - x) (a - b)
+##
+## In the case a = 0, b = 1 we get x (1 - x) which we expect
+##
+## https://en.wikipedia.org/wiki/Logistic_function#Derivative
+dilogit_bounded <- function(x, a, b) {
+  (a - x) * (b - x) * (a - b)
```
M-Kusumgar, line 219: divide here too

M-Kusumgar (changes requested):
Looks neat! i think the maths might be slightly off in one of the things but I could be wrong!

R/sampler-hmc.R
```diff
@@ -0,0 +1,220 @@
...
+    ## Transform from R^n into the model space
+    rn2model = function(theta) {
+      theta[is_semi_infinite] <- exp(theta[is_semi_infinite])
+      theta[is_bounded] <- ilogit_bounded(theta[is_bounded], a, b)
+      theta
+    },
+
+    ## Transform from model space to R^n
+    model2rn = function(x) {
+      x[is_semi_infinite] <- log(x[is_semi_infinite])
```
edknock, line 161: I suggest instead `x[is_semi_infinite] <- log(x[is_semi_infinite] - lower[is_semi_infinite])`. Reasons being: 1. this ensures we transform to R^n and 2. the current version will break if you have a lower bound below 0.

edknock: Relatedly the rn2model would be
```
theta[is_semi_infinite] <- lower[is_semi_infinite] + exp(theta[is_semi_infinite])
```

edknock: And if we wanted to allow semi-infinite but with a finite upper bound we could use `log(upper - x)` as a transform

richfitz: thanks Ed - I've applied the semi infinite fix; I've not done the one for semi infinite in the negative direction yet though; let's wait until we need it first

M-Kusumgar (approved):

PR #14, opened 2024-03-01
Support for combining models

richfitz:
This PR adds support for combining models via adding their log densities.  The basic idea is to support the fairly common case that we have a pair of models, one representing the (log) priors and the other representing the (log) likelihood that we want to join together to make a posterior.  We can then construct these however we want and combine them with `mcstate_model_combine` (or `+` which is syntactic sugar for this). 

The way that different model components combine is described in some detail in the docs but briefly:

* parameters become the union over both models
* domains take the most restrictive per parameter
* densities are summed on the log scale
* gradients are also summed on the log scale, assuming (correctly) zero gradient for variables missing from a component model
* direct sample can only be included from a single model

Because I was finding it hard to work with, I've changed domain slightly to include parameter names.


R/combine.R
```diff
@@ -0,0 +1,196 @@
...
+    name <- "a"
+    model <- a
+  } else {
+    name <- "b"
+    model <- b
+  }
+  if (!all(parameters %in% model$parameters)) {
+    cli::cli_abort(
+      paste("Can't create a direct_sample from these models as '{name}' does",
+            "not contain all parameters in both models"),
```
M-Kusumgar, line 186: im not sure that the name "a" or "b" is useful to the user, i found it actively confusing in the test for this, can we do a
```R
`+.mcstate_model` <- function(x, y,
  names = c(deparse(substitute(x)),deparse(substitute(y))))
```
and pass this down into this function so that we can reference the variable that they tried to sum?

M-Kusumgar (commented):
Nice one, was pretty clear, couple small suggestions!

R/combine.R
```diff
@@ -0,0 +1,203 @@
...
+##' * `gradient`: if *both* models define a gradient, this is the sum
+##'   of the gradients.  If either does not define a gradient, the
+##'   resulting model will not have gradient support.  Set
+##'   `has_gradient = TRUE` within `properties if you want to enforce
+##'   that the combination is differentiable.  If the models disagree
+##'   in their parameters, parameters that are missing from a model
+##'   are assumed (reasonably) to have a zero gradient.
+##'
+##' * `direct_sample`: this one is hard to do the right thing for.  If
+##'   neither models can be directly sampled from that's fine, we
```
weshinsley, line 28:
```suggestion
##'   neither model can be directly sampled from that's fine, we
```

R/combine.R
```diff
@@ -0,0 +1,203 @@
...
+##'   don't directly sample.  If only one model can be sampled from
+##'   *and* if it can sample from the union of all parameters then we
+##'   take that function (this is the case for a prior model when
+##'   combined with a likelihood).  Other cases will be errors, which
+##'   can be avoided by setting `has_direct_gradient = FALSE`in
```
weshinsley, line 33:
```suggestion
##'   can be avoided by setting `has_direct_gradient = FALSE` in
```

R/combine.R
```diff
@@ -0,0 +1,203 @@
...
+##'
+##' @param name_a Name of the first model (defaulting to 'a'); you can
+##'   use this to make error messages nicer to read, but it has no
+##'   other practical effect.
+##'
+##' @param name_b Name of the first model (defaulting to 'b'); you can
+##'   use this to make error messages nicer to read, but it has no
+##'   other practical effect.
+##'
+##' @return A [mcstate_model] object
```
weshinsley, line 62:
```suggestion
##' @return An [mcstate_model] object
```

R/combine.R
```diff
@@ -0,0 +1,203 @@
...
+##' only part that makes a distinction between the two models here;
+##' for all components above they are equivalent.
+##'
+##' @title Combine two models
+##'
+##' @param a The first model
+##'
+##' @param b The second model
+##'
+##' @param properties A [mcstate_model_properties] object, used to
```
weshinsley, line 51:
```suggestion
##' @param properties An [mcstate_model_properties] object, used to
```

weshinsley (approved):
Looks neat - some tiny doc typos!

M-Kusumgar (approved):
Nice one, i like the "lhs" and "rhs", much clearer!

PR #15, opened 2024-03-01
Implement new adaptive sampler

edknock:
This implements an adapted version of Simon Spencer's accelerated shaping and scaling adaptive MCMC algorithm.


R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+    accept <- u < accept_prob
+    if (accept) {
+      state$pars <- pars_next
+      state$density <- density_next
+    }
+
+    internal$iteration <- internal$iteration + 1
+    internal$history_pars <- rbind(internal$history_pars, state$pars) 
+    if (internal$iteration > adapt_end) {
+      internal$scaling_history <- c(internal$scaling_history, internal$scaling)
```
richfitz, line 156: Does this mean that `internal$scaling_history` has length `internal$iteration`? If so can we use this as the source of truth for iteration?

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+    d <- length(internal$mean)
+    internal$autocorrelation <- matrix(0, d, d)
+    internal$vcv <- update_vcv(internal$mean, internal$autocorrelation,
+                               internal$weight)
+    
+    internal$scaling <- initial_scaling
+    internal$scaling_increment <- scaling_increment %||%
+      calc_scaling_increment(length(internal$mean), acceptance_target,
+                             log_scaling_update)
+    internal$n_start <- calc_n_start(acceptance_target)
```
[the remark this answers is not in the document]
richfitz: also what is the '5' from? Would we change that ever? It's not `1 / forget_rate` is it?

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+                     pre_diminish, internal$n_start, log_scaling_update)
+    internal$scaling_history <- c(internal$scaling_history, internal$scaling)
+    internal$autocorrelation <- update_autocorrelation(
+      state$pars, internal$weight, internal$autocorrelation, pars_remove)
+    internal$mean <- update_mean(state$pars, internal$weight, internal$mean,
+                                 pars_remove)
+    internal$vcv <- update_vcv(internal$mean, internal$autocorrelation,
+                               internal$weight)
+    internal$included <- 
+      update_included(internal$included, internal$iteration, is_replacement)
```
richfitz, line 180: I think this would also be easier to understand with the definition here directly

```suggestion
    if (is_replacement) {
      internal$included <- c(internal$included[-1L], internal$iteration)
    } else {
      internal$included <- c(internal$included, internal$iteration)
    }
```

richfitz: and I think you can then fold it in with the earlier `is_replacement` if/else statement, at which point I'll be a bit clearer on what we're doing with this history saver

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+  mcstate_sampler("Adaptive Metropolis-Hastings",
+                  initialise,
+                  step,
+                  finalise)
+}
+
+
+calc_scaling_increment <- function(d, acceptance_target, log_scaling_update) {
+  if (log_scaling_update) {
+    A <- - qnorm(acceptance_target / 2)
```
richfitz, line 200:
```suggestion
    A <- -qnorm(acceptance_target / 2)
```

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  mcstate_sampler("Adaptive Metropolis-Hastings",
+                  initialise,
+                  step,
+                  finalise)
+}
+
+
+calc_scaling_increment <- function(d, acceptance_target, log_scaling_update) {
```
richfitz, line 198: what is 'd' here? It looks like it's a length from the calling?

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+
+calc_proposal_vcv <- function(scaling, vcv, weight, initial_vcv,
+                              initial_vcv_weight) {
+  d <- dim(vcv)[1]
+  
+  weighted_vcv <-
+    ((weight - 1) * vcv + (initial_vcv_weight + d + 1) * initial_vcv) /
+    (weight + initial_vcv_weight + d + 1)
+  
+  2.38 ^ 2 / d * scaling ^ 2 * weighted_vcv
```
richfitz, line 231:
```suggestion
  2.38^2 / d * scaling^2 * weighted_vcv
```

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+                           n_start, log_scaling_update) {
+  scaling_change <- scaling_increment * (accept_prob - acceptance_target) /
+    sqrt(n_start + max(0, iteration - pre_diminish))
+  
+  if (log_scaling_update) {
+    scaling <- pmax(min_scaling, scaling * exp(scaling_change))
+  } else {
+    scaling <- pmax(min_scaling, scaling + scaling_change)
+  }
+
```
richfitz, line 255:
```suggestion
  if (log_scaling_update) {
    pmax(min_scaling, scaling * exp(scaling_change))
  } else {
    pmax(min_scaling, scaling + scaling_change)
  }
```

but are these vectors here (min_scaling, scaling and scaling_change)?

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+    vcv <- autocorrelation - weight / (weight - 1) * qp(mean)
+  } else {
+    vcv <- 0 * autocorrelation
+  }
+  
+  vcv
+}
+
+
+update_included <- function(included, i, is_replacement) {
```
richfitz, line 302: see above, better inlined I think

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+                                 pars_remove)
+    internal$vcv <- update_vcv(internal$mean, internal$autocorrelation,
+                               internal$weight)
+    internal$included <- 
+      update_included(internal$included, internal$iteration, is_replacement)
+
+    state
+  }
+
+  ## Marc; do we want all of these saved out?  Did we ever decide how
```
richfitz, line 185: This comment is now for you Ed!

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+  ## perhaps 'finalise') as I think we've previously collected this up
+  ## at the end of the sampling as the estimated vcv is of interest.
+  internal <- new.env()
+
+  initialise <- function(state, model, rng) {
+    internal$weight <- 0
+    internal$iteration <- 0
+    
+    internal$mean <- state$pars
+    d <- length(internal$mean)
```
richfitz, line 119:
```suggestion
    n_pars <- length(model$parameters)
```

richfitz: and then d becomes n_pars throughout

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+    internal$autocorrelation <- matrix(0, d, d)
+    internal$vcv <- update_vcv(internal$mean, internal$autocorrelation,
+                               internal$weight)
+    
+    internal$scaling <- initial_scaling
+    internal$scaling_increment <- scaling_increment %||%
+      calc_scaling_increment(length(internal$mean), acceptance_target,
```
richfitz, line 126:
```suggestion
      calc_scaling_increment(n_pars, acceptance_target,
```

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+  
+  is_forget_step & is_before_forget_end
+}
+
+
+update_scaling <- function(scaling, iteration, accept_prob, scaling_increment,
+                           min_scaling, acceptance_target, pre_diminish, 
+                           n_start, log_scaling_update) {
+  scaling_change <- scaling_increment * (accept_prob - acceptance_target) /
+    sqrt(n_start + max(0, iteration - pre_diminish))
```
richfitz, line 248: I wonder if passing in `iteration - pre_diminish` might make this easier to understand? (or any part of this subexpression).

richfitz (changes requested):
Some minor thoughts as I try to get my head around this

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+  ## at the end of the sampling as the estimated vcv is of interest.
+  internal <- new.env()
+
+  initialise <- function(state, model, rng) {
+    internal$weight <- 0
+    internal$iteration <- 0
+    
+    internal$mean <- state$pars
+    d <- length(internal$mean)
+    internal$autocorrelation <- matrix(0, d, d)
```
edknock, line 120:
```suggestion
    internal$autocorrelation <- matrix(0, n_pars, n_pars)
```

R/sampler-adaptive.R
edknock: Yes, but also the length of `internal$scaling_history` could be pre-allocated (like you suggest with `internal$history_pars`).

R/sampler-adaptive.R
```diff
@@ -0,0 +1,310 @@
...
+  
+  mean
+}
+
+
+update_vcv <- function(mean, autocorrelation, weight) {
+  if (weight > 1) {
+    vcv <- autocorrelation - weight / (weight - 1) * qp(mean)
+  } else {
+    vcv <- 0 * autocorrelation
```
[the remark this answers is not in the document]
edknock: If `weight == 1` we only have one parameter set so there is no empirical variance/covariance!

R/sampler-adaptive.R
```diff
@@ -0,0 +1,249 @@
...
+    d <- length(internal$mean)
+    internal$autocorrelation <- matrix(0, d, d)
+    
+    internal$scaling <- initial_scaling
+    internal$scaling_increment <- scaling_increment %||%
+      calc_scaling_increment(length(internal$mean), acceptance_target)
+    internal$n_start <- calc_n_start(acceptance_target)
+    
+    internal$history_pars <- c()
+    internal$included <- c()
```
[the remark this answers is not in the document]
edknock: I've changed these

R/sampler-adaptive.R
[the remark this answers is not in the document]
edknock: This is now inline (has become `internal$scaling_weight`). The '5' comes from Simon Spencer's paper, is independent of `1 / forget_rate`.

tests/testthat/test-sampler-adaptive.R
```diff
@@ -0,0 +1,50 @@
+test_that("Empirical VCV calculated correctly", {
```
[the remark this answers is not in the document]
edknock: Done

R/sampler-adaptive.R
edknock: Have now saved only what is useful (and not available elsewhere). We currently return details for multiple chains as just a list of length `n_chains` with the details for each chain, so not combining in any way currently. But wouldn't be clear how to use the VCVs from multiple chains, so something to think about

R/sampler-adaptive.R
```diff
@@ -0,0 +1,308 @@
...
+    if (weight > 2) {
+      autocorrelation <-
+        (1 - 1 / (weight - 1)) * autocorrelation + 1 / (weight - 1) * qp(pars)
+    } else {
+      autocorrelation <- autocorrelation + qp(pars)
+    }
+  }
+  
+  autocorrelation
+}
```
M-Kusumgar, line 286: this function is a little confusing for me, perhaps either a quick chat about it or some comments here, are you doing this `weight > 2` and `weight - 1` everywhere because you've already updated your `weight` before the function call? Also there is more re-scaling here, im assuming to be consistent with the unweighted vcv but perhaps would be nice to just include the rearranged forms of the $\mathbf{\Sigma}_{n}$ that you are using or just using the weighted versions and getting the unweighted version in the finalise function

R/sampler-adaptive.R
```diff
@@ -0,0 +1,308 @@
...
+
+calc_proposal_vcv <- function(scaling, vcv, weight, initial_vcv,
+                              initial_vcv_weight) {
+  n_pars <- nrow(vcv)
+  
+  weighted_vcv <-
+    ((weight - 1) * vcv + (initial_vcv_weight + n_pars + 1) * initial_vcv) /
+    (weight + initial_vcv_weight + n_pars + 1)
+  
+  2.38^2 / n_pars * scaling^2 * weighted_vcv
```
M-Kusumgar, line 240: why is `scaling` squared here? in the algorithm they do not square it

M-Kusumgar (commented):
Nice work getting this all translated to code! I am not super familiar with this so all these suggestions are optional, I may have also misunderstood stuff!

For reference, I have been looking at this picture from the paper as I go through the PR:
anzs12344-fig-0004-m

This is a glossary I am constructing for my understanding but might be useful to anyone else too so putting it here but also please correct me if I have any of them incorrect:
* `mean` = $\overline{\mathbf{X}}_{i}$
* `vcv` = $\mathbf{\Sigma}_{i}$
* `weight` = $n - f(n) + 1$, number of remembered steps
* `n_pars` = $d$
* `initial_scaling` = $\lambda_{start} = \lambda_{0}$
* `scaling` = $\lambda_{i}$
* `initial_scaling_weight` = $n_{start}$
* `scaling_weight` = $n$
* `A` = $A$
* `scaling_increment` = $\delta$
* `acceptance_target` = $a$
* `initial_vcv_weight` = $\nu_{0}$
* $f(n)$ is the number of steps we've forgotten at step $n$

A general comment would be that it would be nice if we follow the outlined algorithm (which I think you mostly do in this PR) but also a small comment wherever you differ from the algorithm and perhaps a short summary as to why. Also perhaps we can derive the `weight` from the forget rate like the algorithm does, the efficiency cost should be minimal but it would be easier to follow the algebra!

R/sampler-adaptive.R
edknock: There is a typo in the algorithm in the paper, it should be squared! It is stated correctly in the text:

image

R/sampler-adaptive.R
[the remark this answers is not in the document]
edknock: We initialise with `weight = 0`, so we hit this bit of code in the end of the first iteration.

You're right that if `forget_rate = 1` it would mean that we would always have one parameter set, it is an extreme choice though! Essentially it would mean we always have no empirical covariance component so the shaping part of the algorithm is entirely non-adaptive. The scaling part would still be adaptive however so it could be useful in case you want to use a fixed VCV but you adaptively scale it?

richfitz (approved):

M-Kusumgar (approved):

PR #16, opened 2024-03-05
Validate initial conditions

richfitz:
This PR copes with two conditions that cause chains not to start correctly:

* the starting point does not have finite probability
* the starting point as sampled by direct_sample falls outside of the model domain

More to do once we have the samplers a bit more sorted out (see monster comment in runner.R) but some basic checks here now which make this nicer to work with in practice.

### DIFF
--- a/R/runner.R
+++ b/R/runner.R
@@ -106,7 +106,7 @@ mcstate_run_chain <- function(pars, model, sampler, n_steps, rng) {
     ## but that has its own issues (e.g., in the heirarchical sampler
     ## we can't extract out the submodel densities, and we might have
     ## issues with rng behaviour of stochastic models).  So this way
-    ## around we'll fail the chain, which is sad if its a chain that
+    ## around we'll fail the chain, which is sad if it's a chain that
     ## only starts running after several hours, and particularly in
     ## parallel we won't find out that anything (and therefore
     ## everything) has failed until everything has completed.
--- a/R/runner.R
+++ b/R/runner.R
@@ -100,6 +100,26 @@ mcstate_run_chain <- function(pars, model, sampler, n_steps, rng) {
   state <- list(pars = pars, density = density)
   sampler$initialise(state, model, rng)
 
+  if (!is.finite(state$density)) {
+    ## Ideally, we'd do slightly better than this; it might be worth
+    ## validing this even earlier (in the parameter validation step)
+    ## but that has its own issues (e.g., in the heirarchical sampler
+    ## we can't extract out the submodel densities, and we might have
+    ## issues with rng behaviour of stochastic models).  So this way
+    ## around we'll fail the chain, which is sad if its a chain that
+    ## only starts running after several hours, and particularly in
+    ## parallel we won't find out that anything (and therefore
+    ## everything) has failed until everything has completed.
+    ##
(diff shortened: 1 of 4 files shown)


M-Kusumgar (approved):
Looks good to me!

R/sample.R
```diff
@@ -270,3 +290,18 @@ restart_data <- function(res, model, sampler, runner) {
        sampler = sampler,
        runner = runner)
 }
+
+
+direct_sample_within_domain <- function(model, rng, max_attempts = 100) {
```
EmmaLRussell, line 295: It's ok for some attempts to fail to be within the domain then?

R/sample.R
```diff
@@ -189,6 +189,26 @@ initial_parameters <- function(initial, model, rng, call = NULL) {
     }
     initial <- matrix(initial, n_chains, n_pars, byrow = TRUE)
   }
+
+  err <- initial < model$domain[, 1] | initial > model$domain[, 2]
```
EmmaLRussell, line 193: This is getting triggered only if initial was provided, since direct_sample_within.. deals with the case where it can't generate in-domain initial itself..?

R/sample.R
```diff
@@ -134,7 +134,7 @@ initial_parameters <- function(initial, model, rng, call = NULL) {
                           arg = "initial", call = environment())
     ## Really this would just be from the prior; we can't directly
     ## sample from the posterior!
-    initial <- lapply(rng, function(r) model$direct_sample(r))
+    initial <- lapply(rng, function(r) direct_sample_within_domain(model, r))
```
EmmaLRussell, line 137: Should max_attempts be configurable here, from a param to initial_parameters?

R/runner.R
```diff
@@ -100,6 +100,26 @@ mcstate_run_chain <- function(pars, model, sampler, n_steps, rng) {
   state <- list(pars = pars, density = density)
   sampler$initialise(state, model, rng)
 
+  if (!is.finite(state$density)) {
+    ## Ideally, we'd do slightly better than this; it might be worth
+    ## validing this even earlier (in the parameter validation step)
+    ## but that has its own issues (e.g., in the heirarchical sampler
+    ## we can't extract out the submodel densities, and we might have
+    ## issues with rng behaviour of stochastic models).  So this way
+    ## around we'll fail the chain, which is sad if its a chain that
```
weshinsley, line 109:
```suggestion
    ## around we'll fail the chain, which is sad if it's a chain that
```

weshinsley (approved):
One more typo!

R/sample.R
richfitz: yeah, I'm not sure about this really. I expect that the user might pass in a function that usually works, but we have to try a few times to get it right and we should sample a couple of times until we hit something reasonable.

R/sample.R
richfitz: yeah, exactly

R/sample.R
richfitz: I hope that it's never needed, tbh. I'm thinking we'll see if anyone ever hits it. But if a model has failed in 100 tries it's never going to succeed I expect and it's mostly there to stop an infinite loop while not bailing too early

EmmaLRussell (approved):

PR #17, opened 2024-03-06
Don't include initial point in chain

richfitz:
This PR drops the initial point from the chain.  It complicates a few things (in implementing support for observers most recently) to retain it so I've dropped it.  Ed indicated that it was useful to know where we've come from so I have retained an `initial` field which holds initial conditions

### DIFF
--- a/R/sample.R
+++ b/R/sample.R
@@ -35,7 +35,25 @@
 ##'   restartable.  This will add additional data to the chains
 ##'   object.
 ##'
-##' @return A list of parameters and densities.
+##' @return A list of parameters and densities; we'll write tools for
+##'   dealing with this later.  Elements include:
+##'
+##' * `pars`: A matrix with as many columns as you have parameters, and
+##'   as many rows as the total number of samples taken across all
+##'   chains (`n_steps * n_chains`)
+##'
+##' * `density`: A vector of model log densities, one per step (length
+##'   `n_steps * n_chains`)
+##'
+##' * `initial`: A record of the initial conditions, a matrix with as
+##'   many columns as you have parameters and `n_chains` rows
+##'
+##' * `details`: Additional details reported by the sampler; this will
+##'   be a list of length `n_chains` (or `NULL`) and the details
+##'   depend on the sampler.  This one is subject to change.
+##'
+##' * `chain`: An integer vector indicating the chain that the samples
+##'   came from (1, 2, ..., `n_chains`)
 ##'
 ##' @export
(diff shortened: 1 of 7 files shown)


R/sampler-hmc.R
```diff
@@ -145,7 +145,7 @@ hmc_transform <- function(domain) {
   }
 
   ## Save finite bounds for later:
-  a <- lower[is_bounded | is_semi_infinite]
+  a <- lower[is_bounded]
```
M-Kusumgar, line 148: Why have we reverted this?

R/sample.R
```diff
@@ -242,6 +269,7 @@ append_chains <- function(prev, curr) {
 
   samples <- list(pars = pars,
                   density = density,
+                  initial = prev$initial,
```
M-Kusumgar, line 272: Should we hold a history of the initial points of each "continue" too? Or is the user expected to hold that if they need it?

M-Kusumgar (commented):
Nice, much neater! Couple small  comments

R/sampler-hmc.R
richfitz: It was causing a warning in the subsetting in the tests, so it wasn't equivalent in the end. We can revisit it later, but I think it was only a theoretical suggested change originally?

R/sample.R
richfitz: The initial points of continue are different; they are within the chain already (the last point of the previous set) and once the chains are merged the idea is that they are indistinguishable from one chain run end to end.  So the (real) initial is the only additional bit of data we need to retain

M-Kusumgar (approved):

R/sample.R
```diff
@@ -202,6 +220,15 @@ combine_chains <- function(res) {
   details <- lapply(res, "[[", "details")
   details <- if (all(vlapply(details, is.null))) NULL else details
 
+  n_pars <- ncol(pars)
+  initial <- vapply(res, "[[", numeric(n_pars), "initial")
```
EmmaLRussell, line 224: This is to wrangle the initial values from each chain into a single matrix?

EmmaLRussell (approved):

weshinsley (approved):
Looks good!

R/sample.R
richfitz: yeah. vapply is "map over a structure and return a 'something' which we'll collect into a structure with one extra dimension". So here numeric(2) would mean we get a 2 x n_chains matrix (or the transpose - I can never remember)

PR #18, opened 2024-03-07
Improvements to sampler initialisation and restart

richfitz:
This PR tidies up how samplers are started:

* e95d9a0 is a small fix that does something I keep wanting - we leave it up to the sampler to return the initial state (so given pars, return a list with pars/density). This is going to be nice once we get observer support and for nested models (with likelihoods that can be separated into different pieces)
* everything else is a larger set of changes that allows saving/loading of internal state from a sampler

After running a chain, in addition to dumping out the rng state we also dump out the sampler state (this is empty for the random walk sampler and just the debug information for the HMC sampler, but for the adaptive sampler in #15 it will be fairly complex).  We pass that back in to `mcstate_run_chain` (along with the rng state) when starting a chain and set that into the sampler after initialisation but before running.

Because the details come from the sampler itself, for appending chains we can just take details from the second sampler which will by then have the state that the combination would have had.

This is going to require a small amount of change to #15 (or this) depending on the merge order.

Looks like the build is broken after merging #15 without a rebase sorry, my fault. This PR is now slightly churny because I zapped trailing whitespace and toned down the number of iteration Ed was using in the tests from 5k to 1k to make it faster.

### DIFF
--- a/R/runner.R
+++ b/R/runner.R
@@ -9,11 +9,12 @@
 ##'
 ##' @export
 mcstate_runner_serial <- function() {
-  run <- function(pars, model, sampler, n_steps, rng) {
+  run <- function(pars, model, sampler, n_steps, rng, sampler_state) {
     lapply(
       seq_along(rng),
       function(i) {
-        mcstate_run_chain(pars[i, ], model, sampler, n_steps, rng[[i]])
+        mcstate_run_chain(pars[i, ], model, sampler, n_steps, rng[[i]],
+                          sampler_state[[i]])
       })
   }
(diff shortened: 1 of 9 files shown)


M-Kusumgar (approved):
Gave it a quick look, seems good to me, would be nice to see this with the adaptive stuff which youre currently working on now :laughing:

M-Kusumgar (approved):
Looks good, tiny comment but doesnt change anything XD

weshinsley (approved):
Great - makes sense and talked through.

PR #20, opened 2024-03-08
Add banana doc

richfitz:
This PR pulls in @MJomaba's banana Rmd doc into the nascent "samplers" vignette. More can be added later!

I've restructured the original a bit to make it a bit faster.  It does make me think that optionally being able to sample many points, or compute many densities, might be nicer.  But it will definitely complicate the interface with unexpected moves between vector and matrix return values.

Closes #11

### DIFF
--- a/R/sampler-adaptive.R
+++ b/R/sampler-adaptive.R
@@ -13,7 +13,7 @@
 ##'
 ##' Our implementation of an adaptive MCMC algorithm is based on an
 ##' adaptation of the "accelerated shaping" algorithm in Spencer (2021).
-##' The algorithm is based on a random-walk Metropolis-Hasting algorithm where
+##' The algorithm is based on a random-walk Metropolis-Hastings algorithm where
 ##' the proposal is a multi-variate Normal distribution centred on the current
 ##' point.
 ##'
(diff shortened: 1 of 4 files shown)


R/sampler-adaptive.R
```diff
@@ -14,7 +14,7 @@
 ##' Our implementation of an adaptive MCMC algorithm is based on an
 ##' adaptation of the "accelerated shaping" algorithm in Spencer (2021).
 ##' The algorithm is based on a random-walk Metropolis-Hasting algorithm where
```
weshinsley, line 16:
```suggestion
##' The algorithm is based on a random-walk Metropolis-Hastings algorithm where
```

vignettes/samplers.Rmd
```diff
@@ -7,4 +7,121 @@ vignette: >
...
+```{r RW_sampling}
+sampler_rw <- mcstate_sampler_random_walk(vcv = diag(2) * 0.01)
+res_rw <- mcstate_sample(m, sampler_rw, 2000)
+plot(res_rw$pars, pch = 19, col = "#ff222277")
+lines(z95[, 1], z95[, 2])
+```
+
+As we can see this is not great.  We can probably improve the samples here by finding a better variance covariance matrix (VCV), but a single VCV will not hold well over the whole surface because it is not very similar to a multivariate normal (that is, the appropriate VCV will change depending on the position in parameter space)
+
+Let's try the Hamilton Monte Carlo (HMC) sampler, which uses the gradient to move efficiently in parameter space:
```
weshinsley, line 118:
```suggestion
Let's try the Hamiltonian Monte Carlo (HMC) sampler, which uses the gradient to move efficiently in parameter space:
```

weshinsley (commented):
Couple of tiny typos...

weshinsley (approved):

PR #21, opened 2024-03-08
Import array utilities

richfitz:
This PR stands in the way of parameter refactoring, and pulls in (with light modification) the array utilities from mcstate.  These make doing some operations with arrays a little easier, but they are a bit of a handful tbh.  We will likely export these later (hence the fairly thorough error handling) but for now I just want them to be able to bind together and expand some arrays

* `array_bind`- I want this for binding arrays together as layers when combining chains (adding a dimension), and end-on-end when appending chains (lengthening an existing dimension)
* `array_reshape` - expands a dimension of an array to a different shape (used to convert a 2d array into a 3d array for example)

Original code here: https://github.com/mrc-ide/mcstate/blob/master/R/arrays.R

See https://github.com/mrc-ide/mcstate2/pull/22 for usage

### DIFF
--- a/R/arrays.R
+++ b/R/arrays.R
@@ -114,6 +114,43 @@ array_reshape <- function(x, i, d, call = NULL) {
 }
 
 
+array_drop <- function(x, i, call = NULL) {
+  dx <- dim2(x)
+  rank <- length(dx)
+  if (rank < max(i)) {
+    cli::cli_abort(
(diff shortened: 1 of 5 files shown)


weshinsley (approved):
OK - all makes sense to me I think - tracking most of the code from mcstate(1)

PR #22, opened 2024-03-08
Store parameters in multidimensional array

richfitz:
This PR updates the samplers to return a matrix of parameters with dimensions `n_pars` x `n_chains` x `n_steps`.  This makes everything much nicer!  However, this will come at the cost of working with the samples, so I added utilities in #21 to help with this, and later we'll need some user-facing ones too I think..  Probably we'll provide some helpful functions for sampling some (or all) of the samples objects and working with the pool, possibly with a burnin/thinning?

### DIFF
--- a/R/runner.R
+++ b/R/runner.R
@@ -13,7 +13,7 @@ mcstate_runner_serial <- function() {
     lapply(
       seq_along(rng),
       function(i) {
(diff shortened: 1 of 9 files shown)


weshinsley (approved):
All makes sense!

PR #23, opened 2024-03-18
Basic observer support

richfitz:
This PR provides the first, inadequately explained, version of sampling with an observer running.  The set up for this is fairly involved because this is not really the sort of thing that people will do outside of the odin-like models probably.

The idea is that we have a model that will remember things about its internal state after it has been run, and that it will be sufficiently costly to run that we want to take advantage of it running just once (or it's stochastic and so we want information that corresponds to the particular realisation that ended up included in the sample).  Most of this is pretty trivial; run the observer after each accepted step, accumulate it at the same point that we save parameters etc.

The complication occurs when thinking about how we create data structures that are nice to work with, so here I've gone with accumulating into arrays with many dimensions in a way that I think is consistent with parameters etc.


tests/testthat/test-observer.R
```diff
@@ -0,0 +1,47 @@
...
+test_that("don't combine things that can't be stuck together", {
+  expect_equal(
+    observer_finalise_auto(list(list(a = 1:5, b = 1:3), list(a = 6:10))),
+    list(list(a = 1:5, b = 1:3), list(a = 6:10)))
+  expect_equal(
+    observer_finalise_auto(list(list(a = 1:5, a = 1), list(a = 6:10, a = 1))),
+    list(list(a = 1:5, a = 1), list(a = 6:10, a = 1)))
+  expect_equal(
+    observer_finalise_auto(list(list(1:5, 1), list(6:10, 1))),
+    list(list(1:5, 1), list(6:10, 1)))
```
M-Kusumgar, line 22: why have we chosen to not output `list(cbind(1:5, 6:10), cbind(1, 1))`? is it because this could be as deep as the user wants and we would have to recursively keep checking dimensions?

tests/testthat/test-observer.R
```diff
@@ -0,0 +1,47 @@
...
+    list(a = cbind(1:5, 6:10)))
+  expect_equal(
+    observer_finalise_auto(list(list(a = 1:5, b = 1:3),
+                                list(a = 6:10, b = 4:6))),
+    list(a = cbind(1:5, 6:10),
+         b = cbind(1:3, 4:6)))
+})
+
+
+test_that("don't combine things that can't be stuck together", {
```
M-Kusumgar, line 13: something like that because it does combine the last one in the test
```suggestion
test_that("combine or don't combine observation correctly", {
```

M-Kusumgar (commented):
Looks good to me! tiny comments, one thing i was wondering about is how strict we should be with the observer finalise and append, you've been fairly lenient by it not erroring when the user has weirdly shaped data but should we enforce nicer shaped observation data? Might be worth having a chat about that type of observation data users are expected to use and if we can find some sort of common structures

M-Kusumgar (approved):
happy after a quick chat with rich about something context to the observer!

tests/testthat/helper-mcstate2.R
```diff
@@ -34,12 +34,29 @@ ex_dust_sir <- function(n_particles = 100, n_threads = 1,
...
+  trajectories <- NULL
+
+  ## In the new dust wrapper we'll need to make this nicer; I think
+  ## that this is pretty painful atm because we wrap via the particle
+  ## filter method in mcstate1.  This version replicates most of what
+  ## we need though, which is some subset of the model
+  details <- function(i, j) {
+    if (save_trajectories) {
+      traj <- trajectories[, i, , drop = FALSE]
+      dim(traj) <- dim(traj)[-2]
```
weshinsley, line 53: Maybe some comments here on what structure `trajectories` is, and what those two lines do / why ? I am staring at it a bit...!

Perhaps better names for i and j - and why don't we use/need j?

weshinsley (approved):
OK - I think this looks good and (mostly) makes sense to me! - perhaps a talkthrough after the break might be good as well.

PR #25, opened 2024-03-20
Correctly restart stochastic models

richfitz:
More bits found while trying to get ready to run observers.

This PR will allow stochastic models to be correctly restarted. This requires that we pull the rng state from the model at the end of running the chain and saving that into the restart data.  This ended up being quite fiddly (81340df) but is not actually a lot of code.  The test in 903e808 failed before this change because the model rng was re-seeded from the chain rng, which is actually pretty terrible as then we reuse a bunch of the same rng chain as before!

With this in, I've made the decision to simplify how model rng setting needs to happen; we now require a raw vector (see the doc change).  This then results in a nice symmetry with getting the rng state at the end of the simulation.

I've then had to change the example a bit, and the validation in model.R has got a bit uglier.  I don't like `mcstate_run_chain2` but don't have any better ideas right now (it's internal only and we can change this whenever).  But very happy to change here

Also notice checks added to the hmc and adaptive samplers that check that they can actually sample from the model provided; these fell through the gaps between PRs earlier.

### DIFF
--- a/R/runner.R
+++ b/R/runner.R
@@ -102,9 +102,9 @@ mcstate_run_chain_parallel <- function(pars, model, sampler, n_steps, rng) {
 
 mcstate_run_chain <- function(pars, model, sampler, n_steps, rng) {
   r_rng_state <- get_r_rng_state()
-  state <- sampler$initialise(pars, model, rng)
+  chain_state <- sampler$initialise(pars, model, rng)
 
-  if (!is.finite(state$density)) {
+  if (!is.finite(chain_state$density)) {
     ## Ideally, we'd do slightly better than this; it might be worth
     ## validing this even earlier (in the parameter validation step)
     ## but that has its own issues (e.g., in the heirarchical sampler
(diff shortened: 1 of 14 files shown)


R/runner.R
```diff
@@ -145,11 +163,15 @@ mcstate_run_chain <- function(pars, model, sampler, n_steps, rng,
   ## surface to the user in the final object (or summarise them in
   ## some particular way with no guarantees about the format).  We
   ## might hold things like start and stop times here in future.
-  internal <- list(used_r_rng = !identical(get_r_rng_state(), r_rng_state),
-                   rng_state = rng$state(),
-                   sampler_state = sampler$get_internal_state())
+  internal <- list(
+    used_r_rng = !identical(get_r_rng_state(), r_rng_state),
+    state = list(
+      chain = state,
```
weshinsley, line 169: Could consider rename `state` to `chain_state` here and on 142 ? Just because of aesthetics of state on 168

weshinsley (approved):
Ok, think I understand this one. 
One tiny readability suggestion

PR #26, opened 2024-03-27
Start on support for nested models

richfitz:
There's more to add here; notably a lot of docs, better examples, better tests and some support for combining nested models, but this is more than big enough to start!

This implements ideas found in both our previous work in mcstate and in drjacoby for working with models where the likelihood can be separated into blocks:

* https://mrc-ide.github.io/mcstate/articles/nested_sir_models.html
* https://mrc-ide.github.io/drjacoby/articles/blocks.html

We introduce a new sampler type (nested random walk) and also add a new model property.  The parameter grouping representation here is designed to be fairly flexible, but we might adjust this later to allow for arbitrary tree shapes of parameters.  For now we assume an optional set of shared parameters, and then a bunch of sets of independent parameters (which need not be the same size per block).  A step will update any shared parameters as a block (if present) then independently for the groups.

I've made a start on some docs, we'll need to update these as things evolve.


R/sampler-nested-random-walk.R
```diff
@@ -0,0 +1,254 @@
...
+  ## There are probably different modes that this could run in, they'd
+  ## be fairly easy to change.  This one would correspond to some sort
+  ## of "full update" mode where everything is done within a step, but
+  ## we could also do one where we allow for picking one update type
+  ## with some schedule or probability and applying that, which would
+  ## allow for faster movement of some part of the chain.  We could
+  ## handle this by additional arguments to the constructor, then
+  ## either changing the behaviour of the step function or swapping in
+  ## a different version.
+  step <- function(state, model, rng) {
```
richfitz, line 116: This method is the only mathematically interesting bit of the PR, really

edknock:
On the whole this is nice, some things I particularly like
- how light the extra parameter setup is for the moment
- how now the parameter groups can be of different dimensions and completely different sets of parameters, whereas before in `mcstate` you had to have the same parameters in each group (and of course the new setup covers that case easily)
- I like the usage of `base` and `groups` - to be honest I was not really a fan of the parameters being `fixed` and `varied` in `mcstate` as to me a "fixed" parameter is the opposite of a "fitted" parameter.

edknock (approved):

tests/testthat/helper-mcstate2.R
```diff
@@ -15,6 +15,57 @@ ex_simple_gamma1 <- function(shape = 1, rate = 1) {
...
+      density = function(x, by_group = FALSE) {
+        z <- dnorm(x, 2^i, 1, log = TRUE)
+        value <- sum(z)
+        if (by_group) structure(value, "by_group" = z) else value
+      },
+      parameter_groups = i)))
+}
+
+
+## This is example is really stupid and does not really help test much
```
weshinsley, line 38:
```suggestion
## This example is really stupid and does not really help test much
```

M-Kusumgar (approved):
Nice one, parameter groups being a positive and increasing sequence definitely helps with the code!

PR #27, opened 2024-04-05
Preprocess DSL input

richfitz:
This PR carves out the fairly tedious bits that need to happen before writing the DSL, in terms of getting user input and getting it ready to process by the parse step (which I'll implement next).  You can see this is already big enough without getting involved in bits that are actually interesting!

I'm not sold on the name `mcstate_dsl` but we can adjust that later perhaps.

See #28 for unmergeable docs describing the approach


R/dsl.R
```diff
@@ -0,0 +1,41 @@
...
+##'
+##' @param x The model as an expression.  This may be given as an
+##'   expression, as a string, or as a path to a filename.  Typically,
+##'   we'll do a reasonable job of working out what you've provided
+##'   but use the `type` argument to disambiguate or force a
+##'   particular interpretation.  The argument uses rlang's quosures
+##'   to allow you to work with expressions directly; see examples for
+##'   details.
+##'
+##' @param type Force interpretation of the type of
```
M-Kusumgar, line 14:
```suggestion
##' @param type Force interpretation of the type of expression.
```

R/dsl.R
```diff
@@ -0,0 +1,41 @@
...
+##' mcstate_dsl("a ~ Normal(0, 1)")
+mcstate_dsl <- function(x, type = NULL) {
+  quo <- rlang::enquo(x)
+  if (rlang::quo_is_symbol(quo)) {
+    x <- rlang::eval_tidy(quo)
+  } else {
+    x <- rlang::quo_get_expr(quo)
+  }
+  exprs <- dsl_preprocess(x, type)
+  NULL
```
M-Kusumgar, line 40: why is this returning null, can we not return exprs to the user?

M-Kusumgar (approved):
Looks good to me, had a suggestion and question but other than that, seems good

R/dsl.R
richfitz: This will eventually return the mode (i.e., something through `mcstate_model()` but that's not implemented yet!)

R/dsl.R
```diff
@@ -0,0 +1,41 @@
...
+##'
+##' @param x The model as an expression.  This may be given as an
+##'   expression, as a string, or as a path to a filename.  Typically,
+##'   we'll do a reasonable job of working out what you've provided
+##'   but use the `type` argument to disambiguate or force a
+##'   particular interpretation.  The argument uses rlang's quosures
+##'   to allow you to work with expressions directly; see examples for
+##'   details.
+##'
+##' @param type Force interpretation of the type of expression.
```
weshinsley, line 14:
```suggestion
##' @param type Force interpretation of the type of expression. 
##' If set, valid strings are `"expression"`, `"text"` or `"file"`
```

weshinsley: Or something like that

weshinsley (approved):
Looks good - one doc suggestion

PR #29, opened 2024-04-05
Allow continuing parallel chains

richfitz:
Small missing feature for the (doomed) parallel runner; continue chains that were started in parallel.

I've folded in an additional small feature which allows the runner to be changed at restart.


weshinsley (approved):
Ok, I understand this one!

M-Kusumgar (approved):
Nice

PR #30, opened 2024-04-05
Add support for progress bars

richfitz:
Create a fairly informative progress bar that shows how each chain, and the overall sampling process, is going.  I've got a small set of vertical bars (one per chain) and a big bar that shows the total sampling progress.  The underlying progress printing machinery will be repurposable for the callr-based parallel runner similar to what we use in mcstate1 at which point it will look slightly more interesting.  I think this is better than one bar per chain in order as the thing that the user cares about is "when is this going to end" and the bar and ETA here capture that more closely.

Code to experiment:

```{r}
options(cli.progress_show_after = 0)
set.seed(1)
m <- ex_dust_sir()
vcv <- matrix(c(0.0006405, 0.0005628, 0.0005628, 0.0006641), 2, 2)
sampler <- mcstate_sampler_random_walk(vcv = vcv)
res <- mcstate_sample(m, sampler, 500, n_chains = 4)
```

producing

progress

Scattered thoughts:

* I'm not 100% certain that this should be an argument to the runner (though it is a property of the runner) and not the sample function, but that's easily changed later and the intent is really that people don't change this very much anyway.
* I've added an Rstudio project
* We depend on global options now (boo) and we need a test setup.R file to stop progress bars making a mess of test output, but I'll get these collected into a doc eventually, similar to what we have in [hipercow](https://mrc-ide.github.io/hipercow/articles/details.html#options)
* there's a bit of noise here with adding assertions etc; there is more to do here on other branches (see #27 for similar code)

### DIFF
--- a/R/progress.R
+++ b/R/progress.R
@@ -1,0 +1,73 @@
+progress_bar <- function(n_chains, n_steps, progress, call = NULL) {
+  progress <- show_progress_bar(progress, call)
+  if (progress) {
+    progress_bar_detail(n_chains, n_steps)
+  } else {
+    progress_bar_null()
+  }
+}
+
+
+show_progress_bar <- function(progress, call = NULL) {
+  if (is.null(progress)) {
+    progress <- getOption("mcstate2.progress", cli::is_dynamic_tty())
+  }
+  ## Error here is not great if we get this from an option; but need
+  ## to disable quoting there in some cases.
+  assert_scalar_logical(progress, call = call)
+}
+
+
+## This will be the case to use where we can report back in detail
+## about what is running (for each chain we can report about the
+## progress within that chain).  The other likely option will be a
+## very coarse detail which is the state of each chain only; that
+## might be the best we can do for some of the other parallel
+## backends, but we'll see how packages that implement progress bars
+## there cope.
+progress_bar_detail <- function(n_chains, n_steps) {
+  e <- new.env()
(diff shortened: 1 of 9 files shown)


R/progress.R
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+}
+
+
+progress_overall <- function(n_chains, n_steps) {
+  if (n_chains == 1) {
+    return(function(n) "")
+  }
+  sym <- unlist(cli::symbol[paste0("lower_block_", 1:8)], use.names = FALSE)
```
weshinsley, line 61: In R from windows command prompt, cli::symbol$lower_block 4 and 8 work; the others do not.
Normal underscore is close to $lower_block_1 but one pixel higher.

You could consider doubling up on ascii codes 176, 177, 178, 219 as a windows-specific best try
░░▒▒▓▓██

In RStudio on Win it is a bit better, but some horizontal spacing variation makes it jiggly

weshinsley (approved):
Looks good - but Windows doesn't have the character set for it - a potential suggestion below!

R/progress.R
richfitz: Another option would be to use the fallback characters? I know cli degrades to something like `_.oO#` which could work.  It might be worth checking to see if this is a font issue too, and/or logging it as an issue in cli as I think the intent is that this should be something that package would smooth over and not us

PR #31, opened 2024-04-09
Parse basic dsl

richfitz:
This PR adds support for parsing the basic DSL, with much more validation to come (and of course, generation).

The DSL so far is an ordered sequence of either assignments with `<-` or distributional statements with `~`; this will feel **very** familiar to the arguments to `quap` in Statistical Rethinking, and that is not a cooincidence.

So the priors corresponding to the regression model from Ch4 that we've been pondering:

```r
a ~ Normal(178, 20)
b ~ Normal(0, 1) # LogNormal distribution not actually supported yet
sigma ~ Uniform(0, 1)
```

The user can assign

```r
a_mean <- 178
a <- Normal(a_mean, 20)
```

and expressions can be used as arguments.  Order matters, so we disallow

```r
a ~ Normal(b, 1)
b <- 10
```

(this is in contrast to odin).

The error handling in the above is quite nice but is _much_ nicer if the code comes from a file

```
> mcstate_dsl("model.R")
Error:
! Invalid use of variable 'b'
→ In expression
  2| a ~ Normal(b, 1)

ℹ 'b' is defined later:
  4| b <- 20
Run `rlang::last_trace()` to see where the error occurred.
```

### DIFF
--- a/R/dsl-parse.R
+++ b/R/dsl-parse.R
@@ -1,0 +1,87 @@
+dsl_parse <- function(exprs) {
+  exprs <- lapply(exprs, dsl_parse_expr)
+
+  type <- vcapply(exprs, "[[", "type")
+  parameters <- vcapply(exprs[type == "stochastic"], "[[", "name")
+  if (anyDuplicated(parameters)) {
+    ## TODO: much better error here.
+    cli::cli_abort("Duplicated parameters")
+  }
+
+  ## TODO: check arguments to calls, perhaps simplest to disallow for now?
+
+  list(parameters = parameters, exprs = exprs)
+}
+
+
+dsl_parse_expr <- function(expr) {
+  if (rlang::is_call(expr, "~")) {
+    dsl_parse_expr_stochastic(expr)
+  } else if (rlang::is_call(expr, c("<-", "="))) {
+    dsl_parse_expr_assignment(expr)
+  } else {
+    dsl_parse_error(
+      "Unhandled expression; expected something involving '~' or '<-'",
+      expr)
+  }
+}
+
+
+dsl_parse_expr_stochastic <- function(expr) {
+  lhs <- expr[[2]]
+  if (!rlang::is_symbol(lhs)) {
+    ## TODO: once we support array expressions this will be relaxed a
+    ## little to allow lhs to be 'symbol[index]'
+    dsl_parse_error("Expected lhs of '~' relationship to be a symbol", expr)
+  }
+  rhs <- expr[[3]]
+
+  ## TODO: this will be derived from source of truth soon:
+  supported <- c("Normal", "Exponential", "Uniform")
+  if (!rlang::is_call(rhs, supported)) {
+    dsl_parse_error(
+      "Expected rhs of '~' relationship to be a call to a distribution")
+  }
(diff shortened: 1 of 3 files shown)


R/dsl.R
```diff
@@ -39,5 +39,18 @@ mcstate_dsl <- function(x, type = NULL) {
     x <- rlang::quo_get_expr(quo)
   }
   exprs <- dsl_preprocess(x, type)
+  dat <- dsl_parse(exprs)
```
weshinsley, line 42: So for now, this is just to test dsl_parse and report errors / success - (dat not used yet)

R/dsl-parse.R
```diff
@@ -0,0 +1,179 @@
...
+  supported <- c("Normal", "Exponential", "Uniform")
+  if (!rlang::is_call(rhs, supported)) {
+    dsl_parse_error(
+      "Expected rhs of '~' relationship to be a call to a distribution",
+      expr)
+  }
+
+  ## This probably requires a little more care in order to know that
+  ## we're not picking up too much or too little.  I'm not sure that
+  ## we can cope with every experssion here too as I think we also
```
weshinsley, line 51:
```suggestion
  ## we can cope with every expression here too as I think we also
```

weshinsley: because typos in temporary comments _really matter_ ....

R/dsl-parse.R
```diff
@@ -0,0 +1,179 @@
...
+        context <- lapply(exprs[name %in% out_of_order], "[[", "expr")
+        names(context) <- sprintf("'%s' is defined later:", out_of_order)
+      } else {
+        ## Could also tell the user about variables found in the
+        ## calling env, but that requires detecting and then passing
+        ## through the correct environment.
+        context <- NULL
+      }
+      ## TODO: It would be nice to indicate that we want to highlight
+      ## the varibles 'err' here within the expression; that is
```
weshinsley, line 172:
```suggestion
      ## the variables 'err' here within the expression; that is
```

R/dsl.R
```diff
@@ -39,5 +39,18 @@ mcstate_dsl <- function(x, type = NULL) {
...
+mcstate_dsl_parse <- function(x, type = NULL) {
+  quo <- rlang::enquo(x)
+  if (rlang::quo_is_symbol(quo)) {
+    x <- rlang::eval_tidy(quo)
+  } else {
+    x <- rlang::quo_get_expr(quo)
+  }
+  exprs <- dsl_preprocess(x, type)
+  dsl_parse(exprs)
+}
```
weshinsley, line 56: What's the difference between mcstate_dsl and mcstate_dsl_parse ?

weshinsley (approved):
All looks clear! One question on the way

R/dsl.R
richfitz: yeah, slowly building up to eventually generating something here

R/dsl.R
richfitz: This one eventually will be something fast to run to parse the expressions and return some metadata but not do any generation; we use that pattern in odin.  It's not really used yet, but will eventually get exported

PR #32, opened 2024-04-15
Better handling of distributions + basic density generation

richfitz:
This PR implements two features that can be separated, but were developed together.  If you would prefer to review them separately I can easily chop them back apart.

First, we add support for some basic distribution objects (e.g., the Exponential distribution or the Normal distribution).  This is mostly just a list with parameters of the distributions and density/sampling functions.  There are two wrinkles here though:

* the domain is not trivial, because for Uniform the practical domain is [`min`, `max`] and not (`-Inf`, `Inf`).  This means that we need the parameters before we can compute the range and that's going to require some thought.  So for now we don't know the domains, which will restrict use with HMC but that's fine as we also don't support differentiation!
* some distributions have multiple parameterisations.  The one that we care about is the negative binomial, which is not implemented here yet.  To keep the PR small I've only used the same three distributions used in the previous version of code - Normal/Exponential/Uniform and so have written this to support the two commonly seen form of Exponential; in terms of rate and in terms of mean.  When we see a call to a distribution (e.g., `Normal(0, 1)` or `Exponential(1)`) we will match the different forms of the distribution with named and positional arguments to try and work out what the user wants.  So `a ~ Exponential(mean = 10)` will select the "mean" form of the exponential distribution while `a ~ Exponential(0.1)` and `a ~ Exponential(rate = 0.1)` select the equivalent distribution parameterised using rate.  Doing this requires some gymnastics with argument matching in dsl-distributions.R

The code generation after this is quite unremarkable I think; note that I have done the density only and not the sampling, though that follows pretty much directly

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -22,3 +22,5 @@
 ^\.github$
 \.*gcov$
 ^_pkgdown\.yml$
+^.*\.Rproj$
+^\.Rproj\.user$
--- a/R/dsl-distributions.R
+++ b/R/dsl-distributions.R
@@ -65,13 +65,13 @@ match_call <- function(args, candidates) {
   }
   detail <- c("x" = sprintf("Failed to match given arguments: %s", given))
 
-  expected <- vcapply(candidates, function(x) paste(names(x), collapse = ", "))
+  expected <- vcapply(candidates, paste, collapse = ", ")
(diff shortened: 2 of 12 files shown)


weshinsley (approved):
Cool - I think I'm keeping up and understanding this! Nothing I can think to add...

PR #33, opened 2024-04-17
Add nested adaptive sampler

edknock:
This is adds `mcstate_sampler_nested_adaptive`, which is to `mcstate_sampler_nested_random_walk` as `mcstate_sampler_adaptive` is to `mcstate_sampler_random_walk`.

This essentially has all the same inputs as `mcstate_sampler_adaptive` but I have extended some (but not all) to be able to vary between `base` and `groups` (and between `groups` too).


R/sampler-nested-adaptive.R
```diff
@@ -0,0 +1,512 @@
...
+##' for all the region-specific parameters, compute the density and then
+##' accept or reject these updates independently.  This is possible
+##' because the change in likelihood in region A is independent from
+##' region B.
+##'
+##' The adaptive proposal algorithm of the non-nested adaptive sampler 
+##' [mcstate_sampler_adaptive] is extended here to adaptively tune the variance
+##' covariance matrix of each of these parameter chunks.
+##'
+##' @title Nested Adaptive Metropolis-Hastings Sampler
```
richfitz, line 21: use inheritParams here

richfitz (changes requested):
Rich will try and integrate this with the changes in mcstate2 and merge today

richfitz (approved):
Merging this now so that we can move with the rename

PR #34, opened 2024-04-25
Drop explicit proposal argument

richfitz:
We don't use this, and it's getting in the way for some new bits of development, so drop it at least for now


weshinsley (approved):
Looks good

PR #35, opened 2024-04-29
Sample simultaneously

richfitz:
~Merge after #34, contains those commits~

This PR adds support for simultaneously sampling from multiple chains at once.  This is triggered by having `pars` be a matrix rather than a vector when passed into a sampler/model.  Models will need to be able to support this and we don't assume that they do; the property `allow_multiple_parameters` can be used to signal this.

This functionality has two purposes:

* to support a generic implementation of parallel tempering, where we will embed a sampler (e.g, random walk, hmc or adaptive walk) within a parallel tempering scheme, requiring us to have each sampler step multiple chains at a single time
* to allow more efficient sampling from simple models (e.g., those that come from the DSL would be easy to generate to support this).

There are lots of changes here, but I think it does all need to come though at once or it loses context. Let's chat through the ideas this week.

Things to note:

* I need to tidy up how `details` is collected at the end of a run; I've not done this even though it's only a few lines
* I do not like `allow_multiple_parameters` as a name, but can't think of a better one
* support for stochastic models is a little hard, and also causes some issues with the wrapper we currently have for dust models; I've punted on that from now and disallow it
* we should be able to support running a fraction of the models only - this needs support in dust

Ignore codecov's comment, it's old and not updating properly


weshinsley (approved):
OK I think I get the gist more or less and we've talked this one through. One documentation comment.

PR #39, opened 2024-07-10
More flexible domain

richfitz:
Less explicit way of specifying the domain.  With this, the user can provide explicit parameter domains for a subset of model parameters and the remainder are assumed to be valid over `(-Inf, Inf)`


weshinsley (approved):
Cool, all makes sense!

PR #40, opened 2024-07-12
Helper function to validate distributions

richfitz:
This PR moves around the distribution validation so that we can reuse it from odin (see https://github.com/mrc-ide/odin2/pull/4).  The error messages are slightly tweaked but the basic structure is the same


R/dsl.R
```diff
@@ -54,3 +54,95 @@ mcstate_dsl_parse <- function(x, type = NULL) {
   exprs <- dsl_preprocess(x, type)
   dsl_parse(exprs)
 }
+
+
+##' Parse an expression as if it was a call to one of mcstate2's
```
weshinsley, line 59:
```suggestion
##' Parse an expression as if it were a call to one of mcstate2's
```

weshinsley: This is excessive grammar pedantry...

R/dsl.R
```diff
@@ -54,3 +54,95 @@ mcstate_dsl_parse <- function(x, type = NULL) {
...
+##' @param expr An expression
+##'
+##' @param name Name for the expression, used in constructing messages
+##'   that you can use in errors.
+##'
+##' @return A list; the contents of this are subject to change.
+##'   However you can (to a degree) rely on the following elements:
+##'
+##' * `name`: The name of the distribution (e.g., `Normal`).  This
+##'   will be the same as the name of the function called in in `expr`
```
weshinsley, line 78:
```suggestion
##'   will be the same as the name of the function called in `expr`
```

R/dsl.R
```diff
@@ -54,3 +54,95 @@ mcstate_dsl_parse <- function(x, type = NULL) {
...
+##'   one is supported.
+##'
+##' * `args`: The arguments that you provided, in position-matched
+##'   order
+##'
+##' * `cpp`: The names of the C++ entrypoint to use.  This is a list
+##'   with elements `density` and `sample` for the log-density and
+##'   sampling functions, and `NULL` where these do not yet exist.
+##'
+##' Currently we also also include:
```
weshinsley, line 90:
```suggestion
##' Currently we also include:
```

weshinsley (approved):
Looks fine! Couple of typos in the docs

PR #41, opened 2024-07-15
Generate sample functions from dsl

richfitz:
So, I thought I'd done more of this but it turns out that I have not done so.

This PR generates `direct_sample` capability for a model, following the same logic as used to generate the density; that is, we work down the statements and evaluate expressions in turn.  The difference is that when computing density when we hit:

```
a ~ Normal(0, 1)
```

we'd save into our final density the result of the density of `density(Normal(a, 0, 1))`, here we need to save into a parameter vector a draw from `Normal(0, 1)`.

To make this easier to implement I've made some changes to the code generation:

* we now use parameter unpacking via the packer; we then work in the "scope" created by that list.  This means we unpack into a list and any assignments would assign into that same list, removing the use of `.x` and the evaluating environment of the function.
* we save the density into a vector as before, but we can just call this `density`
* we do a similar thing with the sample, where we sample into the `env` list and subset out later
* I've written the functions that do the work so that they don't need to know the name of the values used above in the context of the function, even though they are generating code that uses that name

### DIFF
--- a/R/dsl-generate.R
+++ b/R/dsl-generate.R
@@ -77,7 +77,8 @@ dsl_generate_density_stochastic <- function(expr, env, density) {
 
 dsl_generate_sample_stochastic <- function(expr, env) {
   lhs <- bquote(.(env)[[.(expr$name)]])
-  args <- lapply(expr$args, dsl_generate_density_rewrite_lookup, env)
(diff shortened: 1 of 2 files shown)


weshinsley (approved):
Looks good - had to study the syntax a bit here and there...

PR #42, opened 2024-07-17
Implement reflecting boundaries

richfitz:
~Merge after #35, contains those commits~

Only implemented for the non-adaptive algorithm for now.  Ed - can you adapt this for the adaptive one at some point please?

I've added rejection too

### DIFF
--- a/R/sampler-random-walk.R
+++ b/R/sampler-random-walk.R
@@ -31,13 +31,6 @@ mcstate_sampler_random_walk <- function(vcv = NULL, boundaries = "reflect") {
   check_vcv(vcv, allow_3d = TRUE, call = environment())
   internal <- new.env()
 
-  ## We'll also support rejection here but this requires a little work
-  ## in the interface for the multiple-parameter set runner, so that
-  ## we can say "only run this subset please".  We need this for the
-  ## PT algorithms too, in order to not evaluate likelihood for the
-  ## sampling directly from the prior, and this is its own piece of
-  ## work really, which I might start in dust, and then we'll formalise
-  ## as part of the expected model interface.
   boundaries <- match_value(boundaries, c("reflect", "reject", "ignore"))
 
   initialise <- function(pars, model, observer, rng) {
--- a/R/sampler-random-walk.R
+++ b/R/sampler-random-walk.R
@@ -8,9 +8,9 @@
 ##' @param vcv A variance covariance matrix for the proposal.
 ##'
 ##' @param boundaries Control the behaviour of proposals that are
-##'   outside the model domain.  The suported options are:
+##'   outside the model domain.  The supported options are:
 ##'
-##'   * "reflect" (the default): we reflect proposed parmeters that
+##'   * "reflect" (the default): we reflect proposed parameters that
 ##'     lie outside the domain back into the domain (as many times as
 ##'     needed)
 ##'
@@ -33,7 +33,7 @@ mcstate_sampler_random_walk <- function(vcv = NULL, boundaries = "reflect") {
   ## we can say "only run this subset please".  We need this for the
   ## PT algorithms too, in order to not evaluate likelihood for the
   ## sampling directly from the prior, and this is its own piece of
-  ## work really, which I might start in dust, and the we'll formalise
+  ## work really, which I might start in dust, and then we'll formalise
   ## as part of the expected model interface.
   internal$boundaries <-
     match_value(boundaries, c("reflect", "ignore"))
@@ -97,7 +97,7 @@ make_random_walk_proposal <- function(vcv, domain, boundaries) {
 
 
 ## create function to reflect proposal boundaries at pars_min and pars_max
-## this ensures the proposal is symetrical and we can simplify the MH step
+## this ensures the proposal is symmetrical and we can simplify the MH step
 reflect_proposal <- function(x, x_min, x_max) {
   i <- x < x_min | x > x_max
   if (any(i)) {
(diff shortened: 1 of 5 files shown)


R/sampler-random-walk.R
```diff
@@ -7,14 +7,37 @@
 ##'
 ##' @param vcv A variance covariance matrix for the proposal.
 ##'
+##' @param boundaries Control the behaviour of proposals that are
+##'   outside the model domain.  The suported options are:
```
weshinsley, line 11:
```suggestion
##'   outside the model domain.  The supported options are:
```

R/sampler-random-walk.R
```diff
@@ -7,14 +7,37 @@
...
+##'
+##'   * "reflect" (the default): we reflect proposed parmeters that
```
weshinsley, line 13:
```suggestion
##'   * "reflect" (the default): we reflect proposed parameters that
```

R/sampler-random-walk.R
```diff
@@ -7,14 +7,37 @@
...
+##'
+##' The initial point selected will lie within the domain, as this is
+##' enforced by [mcstate_sample].
+##'
 ##' @return A `mcstate_sampler` object, which can be used with
 ##'   [mcstate_sample]
 ##'
 ##' @export
-mcstate_sampler_random_walk <- function(vcv = NULL) {
+mcstate_sampler_random_walk <- function(vcv = NULL, boundaries = "reflect") {
```
weshinsley, line 27: More for thought - wondering why reflection is the default we want, rather than saying, "that one's outside of the domain, let it fail somehow"

R/sampler-random-walk.R
```diff
@@ -7,14 +7,37 @@
...
   check_vcv(vcv, allow_3d = TRUE, call = environment())
   internal <- new.env()
 
+  ## We'll also support rejection here but this requires a little work
+  ## in the interface for the multiple-parameter set runner, so that
+  ## we can say "only run this subset please".  We need this for the
+  ## PT algorithms too, in order to not evaluate likelihood for the
+  ## sampling directly from the prior, and this is its own piece of
+  ## work really, which I might start in dust, and the we'll formalise
```
weshinsley, line 36:
```suggestion
  ## work really, which I might start in dust, and then we'll formalise
```

R/sampler-random-walk.R
```diff
@@ -56,3 +80,53 @@ mcstate_sampler_random_walk <- function(vcv = NULL) {
...
+  x_min <- domain[, 1]
+  x_max <- domain[, 2]
+  function(x, rng) {
+    reflect_proposal(mvn(x, rng), x_min, x_max)
+  }
+}
+
+
+## create function to reflect proposal boundaries at pars_min and pars_max
+## this ensures the proposal is symetrical and we can simplify the MH step
```
weshinsley, line 100:
```suggestion
## this ensures the proposal is symmetrical and we can simplify the MH step
```

weshinsley (approved):
Nice gentle one! Some typo fixes - and one question - I wonder why reflection is what we'd want by default, rather than a sort of "this is out of bounds, hence fails" sort of default that I might have intuitively thought was "strict but fair".

richfitz:
I was going to do "strict but fair" in a different PR because it interacts with some (partially unresolved) issues in the multiple runner.  It's easy to do this in the case where we have a single parameter set being proposed (we'd just return `-Inf`) but where we're working with a bunch simultaneously then we need a standard way of running a subset of a model that can be run in multiple parameter set mode.  I can fold it in here if you'd prefer though

richfitz:
I've added rejection too (see 8ffa375 for the incremental diff), which feels worth having.

As to why reflection by default:

* ignore is the worst option to have, because it might generate things that the user does not want
* reject is inefficient - we spend time throwing up parameters that are impossible. we need to reject if _any_ dimension is out of bounds so this gets worse as we add dimensions
* reflect should be more efficient (see the rejection rate test), and was the default behaviour in mcstate1.

Ed may be able to comment more intelligently than me on this.

edknock:
> * ignore is the worst option to have, because it might generate things that the user does not want
> * reject is inefficient - we spend time throwing up parameters that are impossible. we need to reject if _any_ dimension is out of bounds so this gets worse as we add dimensions
> * reflect should be more efficient (see the rejection rate test), and was the default behaviour in mcstate1.

I agree that ignore is very much "use at your own peril" so clearly should not be default. We only had reflect in mcstate1 so I would favour that as default. It's possible reject is more efficient in some cases (perhaps a reflection could propose implausible values where the reject would at least save the runtime cost of calculating the likelihood?) But overall I think your reasoning re reflect's increasing efficiency is sound enough to have it as default. Plus it might ensure more efficient usage of parallelism in the multiple parameter sets at once case, where the runtime saving for reject only kicks in if you are rejecting all parameter sets?

edknock (approved):

R/sampler-random-walk.R
```diff
@@ -7,14 +7,39 @@
...
   check_vcv(vcv, allow_3d = TRUE, call = environment())
   internal <- new.env()
 
+  ## We'll also support rejection here but this requires a little work
+  ## in the interface for the multiple-parameter set runner, so that
+  ## we can say "only run this subset please".  We need this for the
+  ## PT algorithms too, in order to not evaluate likelihood for the
+  ## sampling directly from the prior, and this is its own piece of
+  ## work really, which I might start in dust, and then we'll formalise
+  ## as part of the expected model interface.
```
edknock, line 40: Do you need this comment now reject has been included as an option?

richfitz: oops, thanks

weshinsley (approved):
Great - let's merge

PR #43, opened 2024-07-18
Helper for expanding domain given a packer

richfitz:
This is something I need for the much-delayed dust/mcstate wrapper, a little tool for expanding out domain entries for logical parameters.  So if we have a packer that tells us that `x` is a vector of length 3 then we want to expand out a domain entry with

```
x: 0, 1
```

to the actual model parameters

```
x[1]: 0, 1
x[2]: 0, 1
x[3]: 0, 1
```

This is surprisingly fiddly to do, but better that we just do it here.  There's some necessary duplication making sure that the domain argument is suitable for use as a domain matrix with the code in `mcstate_model` and over time we might want to harmonise that but nothing there is actually complicated.


weshinsley (approved):
All makes sense - but not sure I know what to do with it yet...

PR #44, opened 2024-07-18
Add domain to models from the dsl

richfitz:
The only tricky bit here was Uniform, and I think this approach is reasonable.


weshinsley (approved):
Looks good

PR #45, opened 2024-07-19
Fix test coverage holes

richfitz:
One unused function, and one bug in an uncovered bit


richfitz:
Merging this one so I can rebase the other work against it @weshinsley

PR #48, opened 2024-07-24
Add utility for tracing random number calls

richfitz:
This PR creates a utility to help me debug an RNG use in https://github.com/mrc-ide/dust2/pull/32, which was ultimately not useful because the RNG use there was in the C level.


weshinsley (approved):
All clear!

PR #49, opened 2024-07-25
Allow simultaneous sampling and boundaries for nested random walk sampler

edknock:


R/sampler-nested-random-walk.R
```diff
@@ -134,16 +182,59 @@ mcstate_sampler_nested_random_walk <- function(vcv) {
...
+              pars_next[i, j] <- state$pars[i, j]
+            }
+          }
+        } else {
+          i <- model$parameter_groups %in% which(!accept)
+          pars_next[i] <- state$pars[i]
+        }
+        ## If e.g. density is provided by a particle filter, would this bit
+        ## mean rerunning it? Increases time cost if so, and would result in
+        ## new value of density (different to that which was accepted)
```
richfitz, line 237: I think it does really, which leaves us in a bit of a pickle.  Options here:

1. forbid this when the model is stochastic
2. allow a model to advertise itself as additive (though some subtleties with the prior here)
3. add a new method that nested models must provide that allows them to replace new groups (in effect moving this part into the model)

I think the latter is the best option, and something we can discuss when we're both back?

tests/testthat/test-sampler-nested-random-walk.R
```diff
@@ -197,3 +202,41 @@ test_that("can run an observer during a nested fit", {
...
+  res1 <- mcstate_sample(m, sampler, 100, n_chains = 3)
+  
+  set.seed(1)
+  runner <- mcstate_runner_simultaneous()
+  res2 <- mcstate_sample(m, sampler, 100, n_chains = 3, runner = runner)
+  expect_equal(res1, res2)
+})
+
+
+test_that("can run nested random walk sampler with rejecting boundaries
```
richfitz, line 224: Probably we need tests of some of the lower level bits in the sampler too here, particularly around the details of rejecting just some samples?

R/sampler-nested-random-walk.R
```diff
@@ -134,16 +182,59 @@ mcstate_sampler_nested_random_walk <- function(vcv) {
...
+            if (!all(i[, j])) {
+              i_group <- model$parameter_groups %in% which(!i[, j])
+              pars_next[i_group, j] <- state$pars[i_group, j]
+            }
+          }
+        } else {
+          i_group <- model$parameter_groups %in% which(!i)
+          pars_next[i_group] <- state$pars[i_group]
+        }
+        density_next <- model$density(pars_next, by_group = TRUE)
```
richfitz, line 210: You can now use index_group to run a subset here

R/sampler-nested-random-walk.R
```diff
@@ -42,12 +42,28 @@
...
+##'     needed)
+##'
+##'   * "reject": we do not evaluate the density function, and return
+##'     `-Inf` for its density instead.
+##'
+##'   * "ignore": evaluate the point anyway, even if it lies outside
+##'     the domain.
+##'
+##' The initial point selected will lie within the domain, as this is
+##' enforced by [mcstate_sample].
```
richfitz, line 60:
```suggestion
##" @inheritParams mcstate_sampler_random_walk
```

richfitz (changes requested):

R/sampler-nested-random-walk.R
richfitz: I've put this down as mrc-5699

R/sampler-nested-random-walk.R
richfitz: actually this needs support from dust too (mrc-5701, mrc-5700)

PR #50, opened 2024-08-01
Import odin1's symbolic differentiation support

richfitz:
Originally from https://github.com/mrc-ide/odin/pull/298 - imported here with minimal changes, except for the beginnings of an exportable interface as we will want to use this from odin2 as well.


weshinsley (approved):
This all looks great! (again!)

PR #52, opened 2024-08-02
Add unique error codes

richfitz:
This PR adds odin-like error codes, following quite closely the implementation in odin2 (see https://github.com/mrc-ide/odin2/pull/5 for details).  There's some harmonisation to do between these later, but I wanted to get this started before it was too big a job later.  Many reasonable models still don't parse of course...


R/dsl-parse-error.R
```diff
@@ -0,0 +1,97 @@
+##' Explain error codes produced by mcstate2.  This is a work in progress,
+##' and we would like feedback on what is useful as we improve it.
+##' The idea is that if you see an error you can link through to get
+##' more information on what it means and how to resolve it.  The
+##' current implementation of this will send you to the rendered
+##' vignettes, but in future we will arrange for offline rendering
```
weshinsley, line 6:
```suggestion
##' vignettes, but in the future we will arrange for offline rendering
```

vignettes/dsl-errors.Rmd
```diff
@@ -0,0 +1,123 @@
...
+
+Example:
+
+```
+a + 1
+```
+
+# `E102`
+
+Invalid left hand side of a relationship operator (`~`).  Currently the left hand side must be symbol, though this will be relaxed in future once we support an array syntax.  However, you may not use things like numbers or function calls on the left hand side.
```
weshinsley, line 38:
```suggestion
Invalid left hand side of a relationship operator (`~`).  Currently the left hand side must be symbol, though this will be relaxed in the future once we support an array syntax.  However, you may not use things like numbers or function calls on the left hand side.
```

vignettes/dsl-errors.Rmd
```diff
@@ -0,0 +1,123 @@
...
+```
+1 <- 2
+f(x) <- g(y)
+```
+
+# `E103`
+
+Your distribution call failed to parse.  This can fail for many reasons, and the details of the failure come from `mcstate2::mcstate_dsl_parse_distribution`
+
+Example reasons for failure include the the rhs being:
```
weshinsley, line 51:
```suggestion
Example reasons for failure include the rhs being:
```

vignettes/dsl-errors.Rmd
```diff
@@ -0,0 +1,123 @@
...
+
+Example:
+
+```
+a <- 1
+b <- 2
+a <- 3 # <= error here
+```
+
+Assignments must be unique within the dsl code because makes it straightforward to trace usage through the dependency graph and from this create a gradient function using automatic differentation.
```
weshinsley, line 89:
```suggestion
Assignments must be unique within the dsl code because this makes it straightforward to trace usage through the dependency graph and from this create a gradient function using automatic differentation.
```

vignettes/dsl-errors.Rmd
```diff
@@ -0,0 +1,123 @@
...
+Example:
+
+```
+a <- 1
+a ~ Normal(0, 1) # <= error here
+```
+
+# `E204`
+
+An assignment (with `<-`) is shaddowing a previous relationship.
```
weshinsley, line 112:
```suggestion
An assignment (with `<-`) is shadowing a previous relationship.
```

vignettes/dsl-errors.Rmd
```diff
@@ -0,0 +1,123 @@
...
+Example:
+
+```
+a ~ Normal(0, 1)
+a <- 10
+```
+
+# `E205`
+
+Variables are used out of order.  If you are using odin this is a big departure - at the moment you must declare your expressions (assignments and relationships) in order.  However, because we forbid multiple assignment we may relax this in the future, but no existing programmes should be changed.
```
weshinsley, line 123:
```suggestion
Variables are used out of order.  If you are using odin this is a big departure - at the moment you must declare your expressions (assignments and relationships) in order.  However, because we forbid multiple assignment we may relax this in the future, but no existing programs should be changed.
```

vignettes/dsl-errors.Rmd
```diff
@@ -0,0 +1,123 @@
...
+
+```
+a <- 1
+b <- 10
+a <- a + b # <= error here
+```
+
+# `E203`
+
+A relationship (with `~`) is shaddowing a previous assignment.  So after assigning to a variable you have declared that the same symbol refers to a parameter.
```
weshinsley, line 101:
```suggestion
A relationship (with `~`) is shadowing a previous assignment.  So after assigning to a variable you have declared that the same symbol refers to a parameter.
```

weshinsley (approved):
All looking good - some trivial typos (which I'll apply and merge)

PR #53, opened 2024-08-06
Expand distribution support

richfitz:
Merge after #51, contains those commits.

Note that this PR will break both dust2 and mcstate2 as they make calls to distributions referenced here, but the fix is easy.

There are more distributions to add - negative binomial and beta binomial are pretty high up - but this fills in some important gaps.

We're also getting to the point where I need to start on overhauling the rng interface generally - the `mcstate_rng` bit was never really meant to be used like this, unfortunately, and we simply don't expose the densities for testing at all.  But a job for a future PR I think


weshinsley (approved):
Talked through - all good!

PR #54, opened 2024-08-07
Report on supported distributions

richfitz:
~Merge after #51, contains those commits.~

A very small helper function that we need for odin2 to know what distribution functions it can use.


R/dsl-distributions.R
```diff
@@ -1,3 +1,29 @@
...
+##' information about which distributions and arguments would succeed
+##' there.
+##'
+##' @title Information about supported distributions
+##'
+##' @return A data.frame with columns
+##'
+##' * `name` the name of the distribution; each name begins with a
+##'   capital letter, and there are duplicate names where different
+##'   parametrisations are supported.
```
weshinsley, line 13:
```suggestion
##'   parameterisations are supported.
```

R/dsl-distributions.R
```diff
@@ -1,3 +1,29 @@
...
+##' * `args` the arguments of all parameters, *except* the random
+##'   variable itself which is given as the first argument to density
+##'   functions.
+##'
+##' We may expand the output here in future to include information on
```
weshinsley, line 18:
```suggestion
##' We may expand the output here in the future to include information on
```

weshinsley: You quite often type "in future" without "the" - so perhaps this is a dialect issue!

weshinsley (approved):
Easy one!

R/dsl-distributions.R
richfitz: hah, I googled that and found *four* variants. I don't have strong feelings though

R/dsl-distributions.R
richfitz: yeah, "in future" reads fine to me, but I can see that "in the future" is probably more Proper

PR #55, opened 2024-08-12
Add some trivial print methods

richfitz:
Lots of room to improve these, but the idea is to help clean up the docs


weshinsley (approved):
These all look look neat!

RStudio underlines them with a dotted line to give the appearance they are clickable, but then the mouse-hover changes to a "no-entry" sign, which is disappointing. Not sure if this is expected, but anyway - one for another time.

PR #56, opened 2024-08-13
Add support for converting to other mcmc types

richfitz:
Similar to #55 in places, but also contains conversion functions for moving from our array format to some other common formats


weshinsley (approved):
Great, all makes sense

PR #58, opened 2024-08-14
Add sign, copysign

richfitz:
As discussed on Tuesday, straightforward implementations here


weshinsley (approved):

weshinsley (commented):
Looks good - I'm a bit confused of where "copysign" fits - the matching odin PR only calls "sign" - but that can wait til we talk later..

PR #60, opened 2024-08-14
High level interface for models

richfitz:
This will be useful in docs and we can guarantee more checks here than using the methods bound into the list


weshinsley (approved):
One typo to redoc - all clear!

PR #61, opened 2024-08-15
Add new functions to derivatives table

richfitz:
This means we now support almost all of the odin functions we assume we can use, which gets us a step closer to being able to write out a public API for that.  I've copied over the implementations for mod/integer division that account for negatives too

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: mcstate2
 Title: Next Generation mcstate
-Version: 0.1.4
+Version: 0.1.5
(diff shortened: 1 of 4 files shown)


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: mcstate2
 Title: Next Generation mcstate
-Version: 0.1.3
+Version: 0.1.4
```
weshinsley, line 3:
```suggestion
Version: 0.1.5
```

weshinsley (approved):
Just the version to bump

PR #66, opened 2024-08-23
Add simple function interface

richfitz:
This PR adds a simple interface that will prevent users having to directly use `monty_model` for the most common case.  I've not used this in `samplers.Rmd` because that model requires a gradient and that's not supported here yet.  I've tidied up the getting started vignette (`monty.Rmd`) which fixes the graph being transposed at least!


R/model-function.R
```diff
@@ -0,0 +1,56 @@
+##' Create a [monty_model] from a function that computes density.
+##' This allows use of any R function as a simple monty model.  If you
+##' need advanced model features, then this interface may not suit you
+##' and you may prefer to to use [monty_model] directly.
```
weshinsley, line 4:
```suggestion
##' and you may prefer to use [monty_model] directly.
```

weshinsley (approved):

PR #69, opened 2024-09-16
Use error parse/explain code from reside.utils

richfitz:
Uses https://github.com/reside-ic/reside.utils/pull/4


weshinsley (approved):
All clear!

PR #72, opened 2024-09-23
callr-based parallel runner

richfitz:
~Merge after #71, contains those commits~

A parallel runner, powered by callr.  Based on the same idea in mcstate


richfitz:
The missing lines here are acceptable, and largely a product of the flakey codecov action - most are unrelated, and we can go through and mop them up later.

There are some missing lines here testing the progress bar bits.  I'm not actually sure how best to do this, and might come up with some gory mocking later once we are happy with the behaviour in practice

weshinsley (approved):
This all runs nicely for me - Cascadia Mono is a good font to use, and somewhere in the docs we should mention that other fonts might not behave nicely.

Also codecov is complaining - covr::report() reports some things we could chase down in a future PR

PR #73, opened 2024-09-26
Packer improvements

richfitz:
The first commit (7aa5a234c22d04cabaa71f2781be6826b393ff7a) fixes a small bug that I noticed when working with dust to unpack output from dust models, where we were treating "scalars" created via the `array` argument slightly differently to those created via the `scalar` argument.

The next couple of commits, and the bulk of this PR fixes a problem I noticed working with this, where we simply don't support packing multidimensional inputs.  I don't think we'll do this very often but it felt like the current implementation was not good enough.  This is surprisingly horrible to do, and I think that's reflected in the error messages and help files.   This is something I feel we'll improve on over time.

### DIFF
--- a/tests/testthat/test-packer.R
+++ b/tests/testthat/test-packer.R
@@ -14,9 +14,11 @@ test_that("trivial packer", {
   expect_error(xp$unpack(1:2),
                "Incorrect length input; expected 1 but given 2")
   expect_equal(xp$pack(list(a = 1)), 1)
-  expect_error(xp$pack(list(a = 1:2)),
-               "Invalid structure to 'pack()'",
-               fixed = TRUE)
+  expect_equal(xp$pack(list(a = 1:2)),
+               1:2)
+  expect_equal(xp$pack(list(a = 1:2, b = 2)),
+               1:2)
+
   expect_equal(xp$index(), list(a = 1))
 })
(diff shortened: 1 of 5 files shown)


tests/testthat/test-packer.R
```diff
@@ -220,3 +220,165 @@ test_that("can't used process with array unpacking", {
...
+  expect_equal(p$unpack(array(1:30, c(2, 3, 5))),
+               list(a = matrix(seq(1, 29, by = 2), c(3, 5)),
+                    b = array(seq(2, 30, by = 2), c(1, 3, 5))))
+})
+
+
+## These tests are the inverse of the tests abvove.
+test_that("Roundtrip scalars stored as zero-length arrays", {
+  p <- monty_packer(array = list(a = integer(0), b = 1L))
+  p$pack(list(a = 1, b = 2))
```
weshinsley, line 241:
```suggestion
  expect_equal(p$pack(list(a = 1, b = 2)), c(1, 2))
```

weshinsley (approved):
OK - I think I've looked at this enough to get it - a couple of doc fixes - 

I am a bit confused in the residual dimensions handling - but I think let's not let that hold up progress - will catch up on it later

PR #74, opened 2024-09-26
Fix doc rendering issues

richfitz:
* missing title tags in the reference
* we would benefit from section headers in the article list

### DIFF
--- a/vignettes/samplers.Rmd
+++ b/vignettes/samplers.Rmd
@@ -34,7 +34,7 @@ m <- monty_example("banana", sigma = 0.5)
 m
 ```
 
-We can plot a greyscale visualisation of its density by computing the density over a grid.  Normally this is not possible of course:
+We can plot a visualisation of its density by computing the density over a grid.  Normally this is not possible of course:
 
 ```{r}
 a <- seq(-2, 6, length.out = 1000)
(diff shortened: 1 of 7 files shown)


vignettes/samplers.Rmd
```diff
@@ -25,53 +25,23 @@ library(monty)
...
+m
 ```
 
 We can plot a greyscale visualisation of its density by computing the density over a grid.  Normally this is not possible of course:
 
 ```{r}
 a <- seq(-2, 6, length.out = 1000)
 b <- seq(-2, 2, length.out = 1000)
-z <- outer(a, b, function(a, b) dnorm(b) * dnorm((a - b^2) / 0.5))
+z <- outer(a, b, function(alpha, beta) {
```
weshinsley, line 42: Above - it's not strictly "greyscale" - more sort of shades of roasted banana from yellow to gold. But it probably doesn't matter...

vignettes/samplers.Rmd
```diff
@@ -104,24 +74,40 @@ points(s[1, ], s[2, ], pch = 19, col = "#00000055")
 
 It is not generally possible to directly sample from a density (otherwise MCMC and similar methods would not exist!).  In these cases we need to use a sampler based on the density and if available possibly the gradient of the density.
 
-We can start with a basic random-walk sampler
+We can start with a basic random-walk sampler:
```
weshinsley, line 77: Above on line 57, a typo - We can check **than** -> **that**

vignettes/samplers.Rmd
```diff
@@ -104,24 +74,40 @@ points(s[1, ], s[2, ], pch = 19, col = "#00000055")
...
-res_rw <- monty_sample(m, sampler_rw, 2000)
-plot(res_rw$pars, pch = 19, col = "#ff222277")
+res_rw <- monty_sample(m, sampler_rw, 1000)
+plot(t(drop(res_rw$pars)), type = "l", col = "#0000ff66",
+     xlim = range(a), ylim = range(b))
 lines(z95[, 1], z95[, 2])
 ```
 
-As we can see this is not great.  We can probably improve the samples here by finding a better variance covariance matrix (VCV), but a single VCV will not hold well over the whole surface because it is not very similar to a multivariate normal (that is, the appropriate VCV will change depending on the position in parameter space)
+As we can see this is not great, exhibiting strong random walk behaviour as it slowly explore the surface (this is over 1,000 steps).  Another way to view this is the parameters varying over steps:
```
weshinsley, line 87:
```suggestion
As we can see this is not great, exhibiting strong random walk behaviour as it slowly explores the surface (this is over 1,000 steps).  Another way to view this is the parameters varying over steps:
```

weshinsley (approved):
Very nice - couple of fixes

PR #76, opened 2024-10-09
Drop observers from monty

richfitz:
We are going to try a new approach in a follow-on PR


weshinsley (approved):
OK - not sure what covr is upset about, but presumably next PR...

richfitz:
yeah - the failure here is mostly because we reduced the denominator, but there are small holes to fill everywhere

PR #77, opened 2024-10-09
Add model-based observer

richfitz:

PR #68, opened 2024-09-05
Implement rerun for random walk samplers

edknock:

PR #78, opened 2024-10-09
Combination of two WIP branches

richfitz:
Merges together #68 and #77 so that Thom can use  it. Don't merge this one!


richfitz:
Closing this now @thomrawson - you should be able to use `main` again

PR #80, opened 2024-10-09
Fixed data for the dsl

richfitz:
As discussed with @edknock, this allows fixed data to be inserted into the DSL code.

This can be reviewed directly, or look at #79 first and then rebase to get a smaller diff.  However, the fallout is not so bad either way


edknock:
This is nice, I think very useful as I imagine people might have pre-analysis tasks that generate hyperparameters, and this seems like a great way to pass them into the dsl.

weshinsley (approved):
All makes sense!

PR #81, opened 2024-10-10
Support for subsetting a packer

richfitz:
We need this in #77 really (or actually in the dust part).

The use case here is that we use a packer for working with trajectories saved when running a model.  We might want to *subset* those trajectories (returning a smaller set than the full) but we also want to be able to use a packer to manipulate these, and also validate that the subset is reasonable.  I've put this all into an experimental packer method for now


weshinsley (approved):
Looks clear!

PR #83, opened 2024-10-11
Drop internal use of 'scalar' in packers

richfitz:
More on packers I am afraid, this time just a tidyup (notice no changes in the tests).  I need this in order to start the multiparameter  packers.

We have an argument of scalar and array, and this PR makes that simply sugar for the user.  Internally we treat everything identically.  If you look at some of the deleted bits you can see we were doing this already really.

### DIFF
--- a/R/packer.R
+++ b/R/packer.R
@@ -277,11 +277,12 @@ monty_packer <- function(scalar = NULL, array = NULL, fixed = NULL,
           i = "Found {length(dups)} duplicate{?s}: {collapseq(dups)}"),
         arg = "scalar")
     }
-    idx[scalar] <- set_names(as.list(seq_along(scalar)), scalar)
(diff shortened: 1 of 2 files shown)


R/packer.R
```diff
@@ -277,11 +277,12 @@ monty_packer <- function(scalar = NULL, array = NULL, fixed = NULL,
           i = "Found {length(dups)} duplicate{?s}: {collapseq(dups)}"),
         arg = "scalar")
     }
-    idx[scalar] <- set_names(as.list(seq_along(scalar)), scalar)
+    shape[scalar] <- rep(list(integer()), length(scalar))
```
richfitz, line 280: This line is basically the entire point of the PR

weshinsley (approved):

PR #85, opened 2024-10-15
Add native SIR filter example

richfitz:
This PR removes `dust` as a dependency and starts the exploration of a proper rng interface.  There's quite a bit in here as a result.  But the main thing that we're **trying** to do is to build a well-behaved stochastic model that can eventually unblock #68.  I don't think that this has everything (you need grouped structure too, right @edknock?) but this gets us most of the way.

The `monty_rng` interface is not suitable for creating models because we can't use the streams in the same way as we do from dust.  If we have `n` values in `size` and `prob` and `n` streams in `rng_state` I want to be able to do `Binomial(size, prob, rng_state)` and have the first stream use the first size and prob, the second stream use the second etc.  This is not implemented in `monty_rng` and this means we have no way to draw random numbers in a way that we can build models that look like dust but without using it.

Here I've sketched out a couple of accessor functions (see the `2` files in both `R/` and `src/`); these allow access in this way.  Then we build a particle filter by hand (this is tedious and was done following the manual implementation in the dust2 helpers: https://github.com/mrc-ide/dust2/blob/main/tests/testthat/helper-dust.R#L5).  With that I have managed to replace all the dust code and replaced it with this model.

The next step will be to make this work with multiple parameter sets, which I'll do in the next PR.

~Merge after #84, contains those commits~

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,7 +31,6 @@ Suggests:
     coda,
     decor,
     knitr,
-    dust,
     mockery,
     mvtnorm,
     numDeriv,
@@ -43,5 +42,3 @@ Suggests:
 Config/testthat/edition: 3
 Language: en-GB
 VignetteBuilder: knitr
-Remotes:
-    mrc-ide/dust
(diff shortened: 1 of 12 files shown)


weshinsley (approved):
Talked through yesterday - all looks good - and we'll mop up the coverage issues a bit later...

PR #88, opened 2024-10-18
Drop references to parameters throughout packer

richfitz:
We use these for state, so parameter is surprising in the docs now.  Also replaced `$parameter` with a function `$names()` so that these become read-only.

This will cause a little breaking change in dust2/odin2 I think but nothing too serious I think.

I need this before complicating the packers a bit further.


R/packer.R
```diff
@@ -1,9 +1,11 @@
-##' Build a parameter packer, which can be used in models to translate
-##' between an unstructured vector of numbers (the vector being
-##' updated by an MCMC for example) to a structured list of named
-##' values, which is easier to program against.  We refer to the
-##' process of taking a named list of scalars, vectors and arrays and
-##' converting into a single vector "packing" and the inverse
+##' Build a packer, which can be used to translate beween an
```
weshinsley, line 1:
```suggestion
##' Build a packer, which can be used to translate between an
```

R/packer.R
```diff
@@ -1,9 +1,11 @@
...
+##' unstructured vector of numbers (the vector being updated by an
+##' MCMC for example) to a structured list of named values, which is
```
weshinsley, line 3:
```suggestion
##' MCMC for example) and a structured list of named values, which is
```

weshinsley (approved):

PR #89, opened 2024-10-21
Add grouped packers

richfitz:
This PR is the first stab at generalising packers to the grouped context; there we have some parameters, some groups and some of the parameters are shared across groups while the others vary.  We might also have fixed data to add in, which might be fixed or varied.


R/packer-grouped.R
```diff
@@ -0,0 +1,243 @@
...
+##' | a  b  c  d1 d2 d3 d4 |
+##' ```
+##'
+##' and **unpack it** into a structured list like
+##'
+##' ```r
+##' list(a = 1, b = 2, c = 3, d = 4:7)
+##' ```
+##'
+##' Our aim here is to the same but to allow some of these values (say
```
weshinsley, line 25:
```suggestion
##' Our aim here is to do the same but to allow some of these values (say
```

R/packer-grouped.R
```diff
@@ -0,0 +1,243 @@
...
+
+  len <- n_shared + n_varied * n_groups
+  nms <- names(p_base$index())
+
+  ## Ideas here for names are:
+  ##
+  ## parameter<group>
+  ## parameter{group}
+  ## group$parameter
+  ## group:parameter
```
weshinsley, line 164: All reasonable - I like the top couple not being literal R syntax...

weshinsley (approved):

PR #98, opened 2024-11-05
Allow using previous samples as initial conditions

richfitz:
This makes picking up after we left off much easier, and also allows the results of a single chain to be used to create starting points for the next chain.  A third way of continuing chains (after continue, continue with `append = FALSE`) but all 3 have their uses I think.


weshinsley (approved):
One typo...!

PR #101, opened 2024-11-06
Add support for thinning samples

richfitz:
Support for postprocessing samples to thin them.  The subset function is one that we might end up exporting in some form later, I;m not sure if this is generally useful.  I did wonder about a "resample" or "subsample" method, possibly with pooling, we can add that here or elsewhere if useful.


edknock (approved):
Looks good to me!

weshinsley (approved):

PR #105, opened 2024-11-07
Allow multiple parameters in models from functions

richfitz:


R/model-function.R
```diff
@@ -22,9 +22,13 @@
...
 ##'   to `monty_packer` instead).
 ##'
+##' @param allow_multiple_parameters Logical, indicating if passing in
+##'   vectors for all parameters will return a vector of densities.
+##'
 ##' @return A [monty_model] object that computes log density with the
 ##'   provided `density` function, given a numeric vector argument
-##'   representing all parameters.
+##'   representing all parameters.  This is `FALSE` by default because
+##'   we cannot determine this automatically.
```
weshinsley, line 31: Think the last sentence should be on the end of `@param allow_multiple_parameters`

weshinsley (approved):
One doc fix...

PR #106, opened 2024-11-08
Proof-of-concept parallel tempering

richfitz:
~Merge after #104, contains those commits~

Closes #24 

A basic parallel tempering sampler.  There's much more to add (particularly embedding other samplers and sorting out sampling from the hot chain), but this is a start at least

This PR does not include tests of statistical correctness; I'm not really sure how we'll do these generally. Marc and I have picked over this pretty closely with some long runs and it's performing well at the moment.

There's quite a lot of refactoring work to do around the samplers once this is in; see comments for details

old comments from a fairly protracted period of validation follow:

----

Marc, a simple example to get you going.  This actually looks pretty good now I think:

```
set.seed(1) # so we see the same thing
pkgload::load_all()
likelihood <- ex_mixture(5)
prior <- monty_dsl({
  x ~ Normal(0, 10)
})
posterior <- likelihood + prior
s <- monty_sampler_parallel_tempering(n_rungs = 10, vcv = matrix(0.1))
res <- monty_sample(posterior, s, 10000, n_chains = 4)

hist(c(res$pars), freq = FALSE)
x <- seq(min(res$pars), max(res$pars), length.out = 1001)
y <- exp(posterior$density(rbind(x)))
lines(x, y / sum(y) / diff(x)[[1]], col = "red")
```

(takes 17s on my machine for 40,000 steps, so 2k steps/s which is ok, especially as this is really 11 times that many samples).

I've run out a much longer chain and this is what I see:

```r
set.seed(42) # so we see the same thing
pkgload::load_all()
likelihood <- ex_mixture(5)
prior <- monty_dsl({
  x ~ Normal(0, 10)
})
posterior <- likelihood + prior
s <- monty_sampler_parallel_tempering(n_rungs = 10, vcv = matrix(0.1))
r <- monty_runner_callr(dust2::dust_openmp_threads(10, "fix"))
res <- monty_sample(posterior, s, 50000, n_chains = 10, runner = r)

hist(c(res$pars), freq = FALSE)
x <- seq(min(res$pars), max(res$pars), length.out = 1001)
y <- exp(posterior$density(rbind(x)))
lines(x, y / sum(y) / diff(x)[[1]], col = "red")
```

(27s for 500,000 points, I can live with that)

image

I suspect we're out by a factor related to the beta swaps at the moment?


R/sampler-parallel-tempering.R
```diff
@@ -0,0 +1,293 @@
...
+##' limit, a normal sampler may only explore a single peak in a model
+##' with many such peaks and will never mix properly.
+##'
+##' @title Parallel Tempering Sampler
+##'
+##' @param n_rungs The number of **extra** chains to run, must be at
+##'   least 1.
+##'
+##' @param vcv The variance covariance matrix for the random walk
+##'   sampler.  You can provide
```
weshinsley, line 45: Sentence to finish

weshinsley (approved):
Looks good - as far as my understanding is stretching... One documentation fix at some point...

PR #112, opened 2024-11-18
Add truncated normal distribution

richfitz:
As requested by Kelly, a truncated normal distribution.  This is actually pretty horrid, and the algorithm here is not universally optimal.  The expressions (needed for the autodiff) are not working because they're really complicated.  But this does mean people can sample from the distributions at least.

Codecov failure is noise due to some templates being counted in the coverage I think


weshinsley (changes requested):
Think I might have found some maths typos...

PR #124, opened 2024-12-06
Port tests to new interface

richfitz:
This ports **most** tests over.

* multinomial tests will require another PR, as the interface there is quite different (multinomial draws take vector args and return vector values)
* the pointer bits still remain, but I might remove that
* the comparison tests with the new interface remain for now

This PR allows @edknock to get the tests set up for Weibull and LogNormal though


weshinsley (approved):
Looks good - a couple of unimportant things I spotted...

PR #127, opened 2024-12-11
Fix generation of multivariate normal random numbers with a single stream

richfitz:
A quite nasty bug triggered in the parallel tempering branch (#106) , which might be the source of the bias there.


R/distributions.R
```diff
@@ -18,46 +22,66 @@ chol_pivot <- function(x) {
 }
 
 
-make_rmvnorm <- function(vcv, centred = FALSE) {
+## There are three uses of this; two use centred and one does not.
+##
+## * monty_example_gaussian uses centrered
```
weshinsley, line 27:
```suggestion
## * monty_example_gaussian uses centred
```

weshinsley (approved):

PR #130, opened 2024-12-13
Add spinner to progress bar

richfitz:
When things don't change that often it looks like things have stalled, a progress bar helps with that.  This PR uses the cli progress bar to add a simple spinner at the start of the progress; I think we had this in mcstate?


R/progress.R
```diff
@@ -155,3 +156,10 @@ with_progress_fail_on_error <- function(progress, code) {
     error = function(e) progress$fail(),
     interrupt = function(e) progress$fail())
 }
+
+
+monty_spinner <- function(date = Sys.Date()) {
+  getOption(
+    "cli.spinner",
+    if (format(date, "%m") == "12") "christmas" else "dots12")
```
weshinsley, line 164: Excellent!

weshinsley (approved):

PR #136, opened 2025-01-31
Fix truncnorm density

edknock:


R/dsl-distributions.R
```diff
@@ -232,6 +232,8 @@ distr_poisson <- distribution(
 distr_truncated_normal <- distribution(
   name = "TruncatedNormal",
   density = function(x, mean, sd, min, max) {
+    dnorm(x, mean, sd, log = TRUE) -
```
richfitz, line 235: Does this need checks to make sure we return -Inf outside of the domain?

richfitz (changes requested):

R/dsl-distributions.R
edknock: Good catch, I've fixed that

richfitz (approved):

PR #139, opened 2025-02-06
Allow domain to be given for DSL models

richfitz:
Ideally we would use truncated distributions, but sometimes this will be preferable

### DIFF
--- a/R/dsl-generate.R
+++ b/R/dsl-generate.R
@@ -27,10 +27,26 @@ dsl_generate <- function(dat) {
 
 dsl_generate_density <- function(dat, env, meta) {
   exprs <- lapply(dat$exprs, dsl_generate_density_expr, meta)
-  body <- c(call("<-", meta[["pars"]], quote(packer$unpack(x))),
+  body_exprs <- c(call("<-", meta[["pars"]], quote(packer$unpack(x))),
             call("<-", meta[["density"]], quote(numeric())),
             exprs,
             call("sum", meta[["density"]]))
+  if (is.null(dat$domain)) {
+    body <- rlang::call2("{", !!!body_exprs)
+  } else {
+    if (nrow(dat$domain) == 1) {
+      domain_min <- dat$domain[[1]]
+      domain_max <- dat$domain[[2]]
+      in_domain <- bquote(x >= .(domain_min) && x <= .(domain_max))
+    } else {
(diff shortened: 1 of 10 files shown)


R/model-function.R
```diff
@@ -84,10 +88,21 @@ monty_model_function <- function(density, packer = NULL, fixed = NULL,
...
+  use_domain <- !is.null(domain)
+  if (use_domain) {
+    domain <- validate_domain(domain, parameters, call = call)
+  }
+
   monty_model(
-    list(parameters = packer$names(),
+    list(parameters = parameters,
          density = function(x) {
+           if (use_domain && any(x < domain[, 1] | x > domain[, 2])) {
```
richfitz, line 101: This needs more care if `allow_multiple_parameters` is `TRUE` and `x` is a matrix

R/dsl-generate.R
```diff
@@ -146,6 +165,13 @@ dsl_generate_domain <- function(dat, meta) {
       }
     }
   }
+
+  ## Same logic as model_combine_domain
```
richfitz, line 169: We might look to do the combination earlier perhaps? Not very straightforward to do so, though

R/dsl-generate.R
```diff
@@ -27,10 +27,26 @@ dsl_generate <- function(dat) {
...
+      domain_max <- dat$domain[[2]]
+      in_domain <- bquote(x >= .(domain_min) && x <= .(domain_max))
+    } else {
+      domain_min <- rlang::call2("c", !!!unname(dat$domain[, 1]))
+      domain_max <- rlang::call2("c", !!!unname(dat$domain[, 2]))
+      in_domain <- bquote(all(x >= .(domain_min) & x <= .(domain_max)))
+    }
+    body <- call(
+      "{",
+      call("if", in_domain, rlang::call2("{", !!!body_exprs), call("{", -Inf)))
```
richfitz, line 48: Note this is the opposite logic to the function case, might be worth inverting it

R/dsl-generate.R
richfitz: no real advantage of doing this actually

weshinsley (approved):
Looks good to me - if you're happy with the comments left when talking this one through!

richfitz:
I fixed them in 8b13833392da82d1905a58b58a6aadb113363684!

PR #144, opened 2025-06-04
Fix two bugs found by Debbie

richfitz:
* `monty_model_function` does not expand the domain
* `call` was not being passed through to the validators, throwing an error while throwing an error


weshinsley (approved):
Looks good!

PR #155, opened 2025-07-30
Add density tests

edknock:
Adding tests for the C++ density functions, truncated normal is outstanding because that possibly needs a bit of a rethink.

Also potentially there needs to be some thought about corner cases on some of the continuous distributions where it could be considered that the distribution becomes a point mass at 0 of infinite density

### DIFF
--- a/tests/testthat/test-densities.R
+++ b/tests/testthat/test-densities.R
@@ -437,13 +437,6 @@ test_that("density::cauchy agrees", {
                density_cauchy(x, location, scale, TRUE))
   expect_equal(dcauchy(x, location, scale, FALSE),
                density_cauchy(x, location, scale, FALSE))
-  
-  ## Corner cases
-  ## dcauchy produces NaN here - do we want that?
-  # expect_equal(density_cauchy(1, 1, 0, TRUE), dcauchy(1, 1, 0, TRUE))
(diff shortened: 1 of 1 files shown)


inst/include/monty/random/density.hpp
```diff
@@ -145,6 +145,8 @@ __host__ __device__ T beta_binomial_ab(T x, T size, T a, T b, bool log) {
   T ret;
   if (x == 0 && size == 0) {
     ret = 0;
+  } else if (x > size) {
+    ret = -random::utils::infinity<T>();
```
edknock, line 149: Ensure -Inf outside domain

inst/include/monty/random/density.hpp
```diff
@@ -288,16 +319,27 @@ __host__ __device__ T weibull(T x, T shape, T scale, bool log) {
...
                 "weibull should only be used with real types");
 #endif
-  const auto ret = monty::math::log(shape) + (shape - 1) * monty::math::log(x) -
-    shape * monty::math::log(scale) - (x / scale)^shape;
+  T ret;
+  if (x < 0) {
+    ret = -random::utils::infinity<T>();
+  } else {
+    ret = monty::math::log(shape) + (shape - 1) * monty::math::log(x) -
+      shape * monty::math::log(scale) - monty::math::pow(x / scale, shape);
```
edknock, line 327: Note here we also correct the power setup at the end

PR #159, opened 2025-08-05
Port HMC to new interface

richfitz:
The next sampler merged over. Straightforward aside from the details processing, which is still quite gross.  In practice this is only the debug information and reporting, and it's only an issue because of the simultaneous runner.  That said, once we get a proper handle on the PT sampler we'll end up using much of the same code there.


edknock (approved):
Looks good to me!

PR #165, opened 2025-08-13
Port parallel tempering sampler to the new interface

richfitz:
This PR ports parallel tempering over to the new format, including state save and restore.  It is the first version of a hierarchical sampler, as the state save/load support makes it fairly easy to reuse one sampler from another.

Unlike the initial version (introduced in #106), this implementation works with any sampler - theoretically at least.  It would be good if we could get this tested and documented in a future PR, but I think the changes should be minimal, really.

I have also, in the interests of clarity, backed out the code that cached the last value of a density calculation in order to decompose the target and base model from each other.  Instead I am assuming that the base model is cheap enough to recompute and deterministic, so we just calculate that on demand.

Also in contrast with the first version, we now actually sample from the prior (or base model) on the hot chain!  To make this efficient we batch compute groups of parameters from the model.  This is likely to have no effect for models that cannot be computed in parallel, but where the density function is expensive and parallel, this implementation should keep all cores busy fairly optimally (the only inefficiency being where `n_rungs` does not divide nicely through the number of steps *and* where the chain is very short so that the extra calculations are wasted).

I have had to add the `control` parameter to the dump, combine and details state function, so that we can access the sub sampler.

For future PRs:

* process the `accept_swap` output to compute an optimal beta (or better, support for doing this adaptively)
* test other samplers, hmc requires a little work for the gradient
* add support for observers
* check if we support stochastic models

None of these features are present on the version on `main`

### DIFF
--- a/R/sampler-parallel-tempering.R
+++ b/R/sampler-parallel-tempering.R
@@ -426,13 +426,12 @@ parallel_tempering_base <- function(model, control, call = parent.frame()) {
     base <- monty_model_split(model, prior_first = TRUE)[[1L]]
   } else {
     base <- control$base
-    ## TODO: we could allow setequal and reorder
-    if (!identical(base$parameters, model$parameters)) {
-      cli::cli_abort("'base' and 'model' must have the same parameters",
-                     call = call)
-    }
   }
-  base
+  if (!setequal(base$parameters, model$parameters)) {
+    cli::cli_abort("'base' and 'model' must have the same parameters",
+                   call = call)
+  }
+  monty_model_reorder(base, model$parameters)
 }
 
 


R/sampler-parallel-tempering.R
```diff
@@ -39,211 +39,328 @@
...
+    list2env(state, env)
   }
 
-  get_internal_state <- function() {
-    keep <- c("accept_swap", "even_step", "last", "state")
-    set_names(lapply(keep, function(k) internal[[k]]), keep)
+  sample <- function(rng) {
+    if (env$index >= size) {
+      env$pars <- direct_sample_many(size, model_base, rng)
+      env$density <- model$density(env$pars)
```
edknock, line 331: if e.g. the model is of form `prior + likelihood` where likelihood is from `dust_likelihood_monty`, does `model$density(env$pars)` involve running the filter/unfilter? so potentially costly?

PR #167, opened 2025-08-15
Add sampler properties

richfitz:
This takes the same approach as model properties, and simplifies checking that the sampler and model are compatible with each other.  This still needs adding to the vignette, but we can sort that later (Ed, perhaps you can look at this while writing up the nested samplers perhaps?)

### DIFF
--- a/man/monty_sampler_properties.Rd
+++ b/man/monty_sampler_properties.Rd
@@ -39,7 +39,7 @@ different to \code{allow_multiple_parameters}, which concerns if the
 \emph{sampler} is able to process multiple parameter sets at once.
 For example, \link{monty_sampler_parallel_tempering} sets
 \code{allow_multiple_parameters} to \code{FALSE} but
-\code{requires_allow_multiple_parameters} to \code{FALSE}, while
+\code{requires_allow_multiple_parameters} to \code{TRUE}, while
 \link{monty_sampler_random_walk} sets the opposite!}
 
 \item{requires_deterministic}{Logical, indicating if the model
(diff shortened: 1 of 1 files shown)


R/sampler.R
```diff
@@ -208,7 +218,93 @@ print.monty_sampler <- function(x, ...) {
 }
 
 
-monty_sampler_state <- function(dump, combine, restore, details,
+##' Describe properties of a sampler.  This is used from
+##' [monty_sampler] to advertise what your sampler does about state,
+##' what it requires from the runner and from the model, so that monty
+##' can ensure that it is only used where it is appropriate.y
```
edknock, line 224:
```suggestion
##' can ensure that it is only used where it is appropriate.
```

R/sampler.R
```diff
@@ -208,7 +218,93 @@ print.monty_sampler <- function(x, ...) {
...
+##' @param requires_gradient Logical, indicating if the model must
+##'   provide a gradient in order to be used with this sampler.
+##'
+##' @param requires_allow_multiple_parameters Logical, indicating if
+##'   the model must be able to accept multiple parameters.  This is
+##'   different to `allow_multiple_parameters`, which concerns if the
+##'   *sampler* is able to process multiple parameter sets at once.
+##'   For example, [monty_sampler_parallel_tempering] sets
+##'   `allow_multiple_parameters` to `FALSE` but
+##'   `requires_allow_multiple_parameters` to `FALSE`, while
```
edknock, line 253:
```suggestion
##'   `requires_allow_multiple_parameters` to `TRUE`, while
```

PR #176, opened 2025-09-26
Support a simple augmented data interface

richfitz:


tests/testthat/helper-monty.R
```diff
@@ -179,3 +179,74 @@ drop_runner <- function(x) {
...
+
+    z <- attr(pars, "data")
+    if (is.null(z)) {
+      z <- z_new
+      density <- density_new
+    } else {
+      ## This recalculates the *old* density, per group.  For an
+      ## expensive model that might not be ideal.
+      density <- likelihood_fn(p, m1, m2, x, z)
+      p_accept <- pmin(1, exp(density_new - density))
```
edknock, line 216: proposal is not symmetric so you need to account for this in the acceptance probability, it should be something like
```
     proposal_ratio <- z * log(p_group_1) + (1 - z) * log(1 - p_group_1) -
        z_new * log(p_group_1) - (1 - z_new) * log(1 - p_group_1)
      p_accept <- pmin(1, exp(density_new - density + proposal_ratio))
```

edknock:
I have set this up so that parallel tempering or HMC do not support augmented data - they could do in future but would need a bit more of a think about how they work.

edknock:
Reopened as a fresh PR (by me rather than Rich) https://github.com/mrc-ide/monty/pull/194

PR #182, opened 2026-02-05
Support arrays in DSL

edknock:
This is a big one! A lot of that is due to all the additional error checks that arrays require.

Basically a lot of ideas for array handling are aligned with odin2. So in this simple model we loop over `i` from 1 to 4 and the parameter names are `c("x[1]", "x[2]", "x[3]", "x[4]").
```
m <- monty_dsl({
  x[] ~ Exponential(i)
  dim(x) <- 4
})
```

With the same index parameters as odin (`i`, `j`, `k`, `l`, `i5`, `i6`, `i7` and `i8`).

As with odin, dimensions of all arrays need to be declared and we can have
```
dim(a) <- dim(b)
```
or
```
dim(a, b) <- ....
```
when `a` and `b` have the same dimensions.

Dimensions must be set upon creation of the model, but do not need to be hardcoded, they can be passed in via `fixed`:
```
m <- monty_dsl({
  x[] ~ Exponential(i)
  dim(x) <- n_x
}, fixed = list(n_x = 4))
```
and fixed parameters can also be arrays themselves.

Multiline equations can be used, but they must be contiguous and cover the whole array in disjoint parts
```
m <- monty_dsl({
  x[1, ] ~ Normal(0, 1)
  x[2:4, ] ~ Normal(x[i - 1, j], 1) 
  dim(x) <- c(4, 5)
})
```
Note that equations are evaluated in the order in which they appear, and that array loops also occur in a natural order so while the above is fine, below is not as e.g. `x[1, 1]` depends on `x[2, 1]` but that is defined subsequently
```
m <- monty_dsl({
  x[1:3, ] ~ Normal(x[i + 1, j], 1)
  x[4, ] ~ Normal(0, 1) 
  dim(x) <- c(4, 5)
})
```

Once parsed and checked, the generated functions for `density` and `direct_sample` are constructed with for loops over array dimensions. Currently we do not generate a `gradient` function for models with arrays - this is doable but fiddly so best kept to a subsequent PR (and it's worth noting that the same is true for odin models currently).


R/dependencies.R
```diff
@@ -0,0 +1,75 @@
...
+          }
+        }
+      }
+    } else {
+      if (is.symbol(e)) {
+        variables$add(deparse(e))
+      }
+    }
+  }
+  descend(expr)
```
M-Kusumgar, line 26: why not just have this function be code? you call it once after defining it

R/sysdata.rda
M-Kusumgar, line 1: what is this file for?

R/dsl-generate.R
```diff
@@ -147,25 +192,61 @@ dsl_generate_density_rewrite_lookup <- function(expr, dest, meta) {
...
+dsl_generate_initialise_arrays_expr <- function(name, arrays, dest, meta) {
+  dims <- arrays$dims[arrays$name == name][[1]]
+  
+  lhs <- bquote(.(meta[[as.character(dest)]])[[.(name)]])
+  
+  if (length(dims) == 1) {
+    expr <- call("<-", lhs, call("numeric", dims[[1]]))
+  } else {
+    expr <- call("<-", lhs, call("array", 0, rlang::call2("c", !!!dims)))
+  }
```
M-Kusumgar: why is this distinction necessary? isnt `array(0, c(5))` is the same as `numeric(5)`?

R/dependencies.R
```diff
@@ -0,0 +1,75 @@
...
+  variables <- collector()
+  descend <- function(e) {
+    if (is.recursive(e)) {
+      if (rlang::is_call(e) && is.symbol(e[[1]])) {
+        functions$add(deparse(e[[1L]]))
+      }
+      is_dim_read <- rlang::is_call(e, c("length", "dim")) &&
+        length(e) == 2 && is.symbol(e[[2]])
+      if (is_dim_read) {
+        variables$add(dsl_parse_dim_name(deparse(e[[2]])))
```
weshinsley, line 12: This is the only line different from odin2::dependencies.R - but I can't really think of a nice way of dealing with the dupliaction. On the plus side, odin2 has already been reviewed, so this is all good!

R/dsl-parse.R
```diff
@@ -35,61 +44,749 @@ dsl_parse_expr <- function(expr, call) {
...
+      dsl_parse_error(
+        "Invalid use of unary minus in lhs of array assignment",
+        "E115", expr, call)
+    }
+    err <- intersect(INDEX, ret$depends$variables)
+    if (length(err) > 0) {
+      dsl_parse_error(
+        paste("Invalid use of special variable{?s} in lhs of array",
+              "assignment: {squote(err)}"),
+        "E116", expr, call)
```
M-Kusumgar, line 354: ls it worth letting people know about the implicit order of the indices on the lhs? like it is always i, j, k, l, ...

M-Kusumgar (commented):
Havent looked at it super carefully but have looked at it. Had a couple of minor comments but on the whole code looks good!

R/sysdata.rda
weshinsley: This is the sort-of-compiled help information - you generate it with `scripts/update_parse_errors` whenever you change `vignettes/errors.Rmd`. The `sysdata.rda` file is special and gets loaded into the environment when you call `library` somehow (I forget how...) 

So
```
library(odin2)
odin_error_explain("E1006")
```

will get you some text, even though `odin_error_explain` appears on the surface to use an unknown variable `errors`

```
odin_error_explain <- function(code, how = "pretty") {
  error_explain(errors, code, how)
}
```

R/dependencies.R
weshinsley: If I understand the question... it's recursive so calls itself from within with different args. So you need the inner func, and this call kicks off at the top level.

weshinsley (approved):
Looks pretty good as far as I can see Ed - lots of it exists in odin2, but it's different enough that I don't really see a nice way of factoring it differently. And on the plus side, the odin2 it comes from has been live and tested for a while...

R/sysdata.rda
M-Kusumgar: ahhh okay makes sense!

R/dependencies.R
M-Kusumgar: i missed the recursive branch!

R/dependencies.R
edknock: Yeah as you can tell the whole file is lifted from odin2 with that minor change, and monty can't depend upon odin2 so need them here (and not sure they are functions that need to be exported really)

R/dsl-parse.R
edknock: This error is about not using any of i, j, k, l etc on the lhs

https://github.com/mrc-ide/monty/blob/9a81669eeb7300f4e02de43d60e8d577ab92d85c/vignettes/dsl-errors.Rmd#L212-L224

there is the separate error about incorrect use of i, j, k, l etc on the rhs

https://github.com/mrc-ide/monty/blob/9a81669eeb7300f4e02de43d60e8d577ab92d85c/R/dsl-parse.R#L392-L396

https://github.com/mrc-ide/monty/blob/9a81669eeb7300f4e02de43d60e8d577ab92d85c/vignettes/dsl-errors.Rmd#L126-L144

Is that enough? I'll add some explanation to the DSL vignette in separate PR about how to use arrays, that's probably a good place to explain about using index variables correctly

R/dsl-generate.R
edknock: Good point, while not strictly identical according to R, they are functionally the same wrt to usage here so I'm removing the need to treat the 1d case separately

M-Kusumgar (approved):
awesome! looks good

PR #183, opened 2026-04-22
DSL array fixes

edknock:
Currently this throws an (uninformative) error

```
fixed = list(n_beta = 5)
m <- monty_dsl({
  beta[1] ~ Exponential(mean = 0.3)
  beta[2:n_beta] ~ Exponential(mean = beta[i - 1])
  dim(beta) <- n_beta
}, fixed = fixed)
```

this is due to the usage of the fixed parameter `n_beta` in indexing on the LHS. The fixes here now allow this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: monty
 Title: Monte Carlo Models
-Version: 0.4.10
+Version: 0.4.11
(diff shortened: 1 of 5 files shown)


weshinsley (approved):
Looks good to me - think I'll need to study it a bit more to properly get the flow into my brain - but let's get this merged so you can get on...

PR #184, opened 2026-04-28
Add save_full_chains option

edknock:
In mcstate when thinning (and/or discarding burnin), the full chains of the parameters and densities were retained (useful for running diagnostics or making traceplots).

With `monty` there are two ways to thin:
1. by setting `burnin` and/or `thinning_factor` in `monty_sample` (so you thin as you go along)
2. by thinning after running using `monty_samples_thin`

Neither method allowed individuals to retain the full chains of the parameters. With the mpox model I would use method 2 and separately save the full chains of parameters before thinning (the main purpose of thinning in this case was to cut down on file size, most of which is due to the trajectories which are typically make for a much larger object than the parameters array).

The issue with that approach is if you run chains long enough you can run out of memory - thinning as you go along solves this problem but at the moment it is impossible to recover the full chains.

Adding this option to both thinning methods makes this easier in general for users.

Some code to try

```
m <- monty_dsl({
  a ~ Exponential(1)
  b ~ Normal(0, 1)
})

sampler <- monty_sampler_hmc()


## running without thinning
set.seed(1)
res1 <- monty_sample(m, sampler, 1000, n_chains = 4)


## running with thinning but not saving full chains
set.seed(1)
res2 <- monty_sample(m, sampler, 1000, n_chains = 4,
                     burnin = 500, thinning_factor = 2)
identical(res2$full_chains, NULL)


## running with thinning, saving full chains
set.seed(1)
res3 <- monty_sample(m, sampler, 1000, n_chains = 4,
                     burnin = 500, thinning_factor = 2, save_full_chains = TRUE)
identical(res3$full_chains$pars, res1$pars)
identical(res3$full_chains$density, res1$density)


## thinning after running, results will be same as thinning while running
res4 <- monty_samples_thin(res1, burnin = 500, thinning_factor = 2)
identical(res2, res4)

res5 <- monty_samples_thin(res1, burnin = 500, thinning_factor = 2,
                           save_full_chains = TRUE)
identical(res3, res5)

## full_chains is a monty_samples object so can be converted with
## posterior::as_draws_df (for e.g. diagnostics)
res1_df <- posterior::as_draws_df(res1)
res3_df <- posterior::as_draws_df(res3$full_chains)
identical(res1_df, res3_df)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: monty
 Title: Monte Carlo Models
-Version: 0.4.11
+Version: 0.4.12
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Wes", "Hinsley", role = "aut"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -15,7 +15,7 @@ Description: Experimental sources for the next generation of mcstate,
 License: MIT + file LICENSE
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.3
 URL: https://mrc-ide.github.io/monty, https://github.com/mrc-ide/monty
 BugReports: https://github.com/mrc-ide/monty/issues
 Imports:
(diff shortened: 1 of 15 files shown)


weshinsley (approved):
Cool - looks good to me Ed, think I can follow it all!

PR #185, opened 2026-05-08
Rename grouped parameters and fix index function

### LINKED ISSUE #141: Index incorrectly setup for grouped packer
I found an issue with how the grouped packer sets up the `index`. This is causes issues with the mpox model.

Suppose we have the following packer

```
process <- function(res) {
  list(z = res$x + res$y)
}

p <- monty_packer(c("x", "y"), fixed = list(w = 1), process = process)
```

only the fitted parameters `x` and `y` appear in the index, and `w` and `z` do not:

```
> p$index()
$x
[1] 1

$y
[1] 2
```

Now if we try a similar grouped packer
```
p_group <- monty_packer_grouped(c("a", "b"), c("x", "y"), fixed = list(w = 1),
                                shared = "y", process = process)
```
the index incorrectly includes `w` and `z`
```
> p_group$index()
$a
$a$x
[1] 2

$a$y
[1] 1

$a$w
[1] 1

$a$z
[1] 3


$b
$b$x
[1] 3

$b$y
[1] 1

$b$w
[1] 1

$b$z
[1] 4
```

The `index` is setup here https://github.com/mrc-ide/monty/blob/main/R/packer-grouped.R#L227 by unpacking a vector of sequenced integers. This is causes additional problems in the case where if a parameter is out-of-bounds unpacking would throw an error - this inevitably happens here if e.g. you have a parameter that can only take values in (0,1) and an integer value will lead to an error as out-of-bounds.

edknock:
This is starting point of dealing with grouped/nested models by dealing with some largely internal changes to the packer. Note that this merges into groups-main, which we will ultimately merge into main.

The key point is changing how grouped parameters will be named. We have a grouped packer already on main. Suppose we have groups `a` and `b` and then we have three parameters in each group `alpha`, `beta` and `gamma` where the first two are scalar and `gamma` is a 2x2 matrix. Suppose further that `beta` is shared across the two groups (it has the same value across the two groups) while the others vary between groups. We set the packer up as

```
packer <- monty::monty_packer_grouped(groups = c("a", "b"),
                                      scalar = c("alpha", "beta"),
                                      array = list(gamma = c(2, 2)),
                                      shared = "beta")
```
which leads to parameter names
```
> packer$names()
 [1] "beta"          "alpha<a>"      "gamma[1,1]<a>" "gamma[2,1]<a>" "gamma[1,2]<a>" "gamma[2,2]<a>"
 [7] "alpha<b>"      "gamma[1,1]<b>" "gamma[2,1]<b>" "gamma[1,2]<b>" "gamma[2,2]<b>"
 ```

However for groups we cannot use something like `<group>` as syntax in the DSL. In the DSL we will write
`alpha | group` to signify a parameter to vary between groups. So here I change the names of the parameters to be consistent with that
```
> packer$names()
 [1] "beta"           "alpha | a"      "gamma[1,1] | a" "gamma[2,1] | a" "gamma[1,2] | a"
 [6] "gamma[2,2] | a" "alpha | b"      "gamma[1,1] | b" "gamma[2,1] | b" "gamma[1,2] | b"
[11] "gamma[2,2] | b"
```

The other bit in this PR is some fixing to the index function in the packer where the index is supposed to just include non-fixed parameters but currently it includes fixed parameters e.g.
```
> packer <- monty::monty_packer_grouped(groups = c("a", "b"),
+                                       scalar = c("alpha", "beta"),
+                                       fixed = list("delta" = 1),
+                                       shared = "beta")
> packer$index()
$a
$a$alpha
[1] 2

$a$beta
[1] 1

$a$delta
[1] 1


$b
$b$alpha
[1] 3

$b$beta
[1] 1

$b$delta
[1] 1
```    
with the fix we now get
```
> packer$index()
$a
$a$alpha
[1] 2

$a$beta
[1] 1


$b
$b$alpha
[1] 3

$b$beta
[1] 1
```
which deals with https://github.com/mrc-ide/monty/issues/141

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: monty
 Title: Monte Carlo Models
-Version: 0.4.11
+Version: 0.4.12
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Wes", "Hinsley", role = "aut"),
--- a/R/packer-grouped.R
+++ b/R/packer-grouped.R
@@ -224,8 +224,19 @@ monty_packer_grouped <- function(groups, scalar = NULL, array = NULL,
     x
   }
 
-  idx <- unpack(seq_len(len))
-
+  #idx <- unpack(seq_len(len))
+  
+  ## This is a simplified version of unpack above to create index
+  base <- c(set_names(vector("list", length(nms)), nms))
+  base[shared] <- d_shared$packer$unpack(d_shared$index_packed)
+  idx <- rep(list(base), n_groups)
+  x_varied <- matrix(d_varied$index_packed, n_varied, n_groups)
+  for (i in seq_len(n_groups)) {
+    idx[[i]][varied] <- d_varied$packer$unpack(x_varied[, i])
+  }
+  names(idx) <- groups
+  
+  
   ## This would be really quite hard, and I'm not sure how effective?
   ## But we might want it later to support multistage simulation in
   ## dust2.
--- a/R/packer-grouped.R
+++ b/R/packer-grouped.R
@@ -224,8 +224,6 @@ monty_packer_grouped <- function(groups, scalar = NULL, array = NULL,
     x
   }
 
-  #idx <- unpack(seq_len(len))
-  
   ## This is a simplified version of unpack above to create index
   base <- c(set_names(vector("list", length(nms)), nms))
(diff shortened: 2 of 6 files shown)


M-Kusumgar (commented):
LGTM, just confirming that the bit with the index is just giving the shared parameters the same indices but not shared parameters different indices in different groups?

M-Kusumgar (approved):
sorry meant to approve

PR #186, opened 2026-05-08
Combine groups

edknock:


edknock:
This now being handled in https://github.com/mrc-ide/monty/pull/190

PR #187, opened 2026-05-15
Add groups into the DSL

edknock:
This is the basic setup of groups in the DSL.

In this example

```
fixed = list(n = 3, mean_beta = c(-1, 0, 1), a = list(z = 2), b = list(z = 3))
m <- monty_dsl({
      alpha ~ Exponential(mean = 3)
      beta[] ~ Normal(mean_beta[i], 1)
      lambda[1:2] <- 2 * i
      lambda[n] <- 2 * i + 1
      gamma | region <- alpha * z
      x[1] | region ~ Exponential(lambda[i] + gamma)
      x[2:n] | region ~ Exponential(lambda[i]^2 + gamma)
      dim(x, lambda, beta, mean_beta) <- n
      region <- group()
    }, groups = list(region = c("a", "b")), fixed = fixed)
```

- `region` is our grouping variable 
- when something is defined without `|` on the lhs this means it is shared across all groups, so `alpha`, `beta` and `lambda` are all shared across groups
- when something is defined with `|` on the lhs this means it is group-specific (essentially we iterate over groups), so `gamma` and `x` are region-specific
- the `fixed` list contains fixed data that are shared along with lists of group-specific fixed data
- grouped parameters can depend upon shared parameters but not vice versa and grouped parameters cannot depend upon grouped parameters from another group
- there is no use of `|` on the rhs - it is implicit whether things are shared or grouped given how they are defined in their own equation

Currently a single density is produced by the `density()` function, in a subsequent PR I will add the breakdown by group (and the shared component).

Beyond checking code, I would appreciate thoughts on syntax here! Particularly:

- Do we think `region <- group()` is fine? I might need to make `group` a protected word in `monty` which is maybe not great if people want to use `group` as the grouping variable? Could go with `region <- grouping()` if so?
- Is the structuring of `fixed` re shared fixed data and grouped fixed data fine? It matches how fixed data is setup for a grouped packer.


R/dsl-generate.R
```diff
@@ -145,47 +171,94 @@ dsl_generate_density_stochastic <- function(expr, meta) {
...
+                                                           has_groups))
     
   if (!is.null(array)) {
     expr <- dsl_generate_array_loops(expr, array, meta)
   }
+  if (has_groups) {
+    if (is_grouped) {
+      expr <- dsl_generate_group_loops(expr, group_data$groups, meta)
+    } else {
+      expr <- dsl_generate_group_loops(expr, group_data$groups[1], meta)
```
M-Kusumgar, line 192: this feels like a weird choice, or maybe im misunderstanding, are you using the first group as a "default" choice here? i ran it up with a test example and seems like youre extracting `pars[["a"]][["alpha"]]` but `alpha` doesnt fall into group `a`

M-Kusumgar (commented):
the code largely looks good! just have a question and a couple other tiny ones for my understanding so lets chat when we are both in?

M-Kusumgar (approved):
thanks for the chat, looks good!

